PR #30, opened 2018-05-20
Exit and logout button

AntiSC2:
This pull request adds a new exit button at the bottom of the menu and it takes the user back to their rooms. A new logout button was also added to the rooms page that uses the existing logout function on the server side. I was not quite sure about the placement of the logout button but I think it looks fine now and I also added a placeholder button that in the future can be used to alter account settings like changing passwords.

screenshotexitbutton
logout-button


Kruptein:
Looks good, I'll check the actual changes later

Issue #47, opened 2018-07-06
Ability to create notes


AntiSC2:
Right now you can create a note by creating an empty token but it would be nice if this was more officially supported so that both players and the DM would be able to create notes.

- [ ] Have a note section in the accordion menu
- [ ] See list of all notes created there and have a button to create new ones
- [ ] Ability to leave a note on a grid
- [ ] When that note is clicked on it will show dialog with the actual text inside

Kruptein:
Seems useful.

The addition to show a note to all players as a DM is also something that I would like to add, this functionality already exists for assets in a limited way in a separate branch currently.

One thing I do like in the current system is that I'm able to draw boxes with a transparent fill over a room and give it an annotation so that when I mouse over the room I have all the info on the top of the screen for that room, it's pretty handy!  But I do agree that notes on the grid have their use as well.

AntiSC2:
@Kruptein, I have started working on this so you can assign me for now

AntiSC2:
#49

Kruptein:
FYI I've ported the notes system to the new Vue UI system in the vue branch, which will be merged in dev relatively soon.

The main file regarding notes is now in `game/ui/note.vue` and a small part is in `game/ui/menu/menu.vue`

Kruptein:
This has been superseded by the overhauled notes UI a couple of releases ago.

PR #49, opened 2018-07-11
Implement basic note taking system

AntiSC2:
This implements a basic note taking system, both on the client side and the server side. In the side menu you can now find a note section where you can create notes and see all the notes that belongs to you. You can write in them, rename them and delete them and all of this is also saved by the server. Because the server now saves notes the save version is also bumped to 2 as old saves won't function correctly.


AntiSC2:
How the menu looks:
notetakingsystem
The dialog:
note2

Kruptein:
Is this fully completed or still work in progress?

AntiSC2:
The feature that is currently missing that was mentioned in #47 is the ability to place them on the grid. Everything else is working right now, creating, editing, renaming, deleting, saving and loading. The reason the commit was amended was because I rebased it after the #51 fix.

Have been a bit busy later so have not started to work on placing notes on the grid, so I would have no problem if you want to merge this right now or if you want, I can also implement the missing feature in this PR.

PlanarAlly/planarally.py
```diff
@@ -148,6 +149,31 @@ def get_active_location(self, username):
     def add_new_location(self, name):
         self.locations[name] = Location(name, self)
 
+    def add_new_note(self, data, username):
+        self.notes
```
Kruptein: This line seems redundant?

PlanarAlly/planarally.py
```diff
@@ -148,6 +149,31 @@ def get_active_location(self, username):
...
+        user_notes = []
+        for uuid, note in self.notes.items():
+            if note[0] == username:
+                user_notes.append({"uuid": uuid, "name": note[1], "text":note[2]})
+        return user_notes
+
+    def delete_note(self, uuid, username):
+        if self.notes[uuid][0] == username:
+            del self.notes[uuid]
+            print("Working!")
```
Kruptein: Get rid of this print ;)

PlanarAlly/planarserver.py
```diff
@@ -411,6 +411,48 @@ def shape_wrap(player, shape):
     await sio.emit("set gridsize", grid_size, room=location.sioroom, skip_sid=sid, namespace="/planarally")
 
 
+@sio.on("new note", namespace="/planarally")
```
Kruptein: This event name is not consistent with the other events you created.

This is something that I've not been consistent with myself and want to cleanup later, so this is not super important to immediately fix.

ts_src/manager.ts
```diff
@@ -125,6 +129,32 @@ export class GameManager {
         Settings.board_initialised = true;
     }
 
+    newNote(name: string, text: string, show: boolean, sync: boolean, uuid:string = ""): void {
```
Kruptein: I have a slight preference to using 

```typescript
newNote(name: string, text: string, show: boolean, sync: boolean, uuid:string = ""): void {
    uuid = uuid || uuidv4();
```

ts_src/manager.ts
```diff
@@ -125,6 +129,32 @@ export class GameManager {
...
+
+        const note = new Note(name, text, uuid);
+
+        if (sync) {
+            socket.emit("new note", note.as_dict());
+        }
+
+        this.notes.set(uuid, note);
+
+        let button = $('<button>' + note.name + '</button>');
```
Kruptein: This can be changed from `let` to `const`

ts_src/note.ts
```diff
@@ -0,0 +1,95 @@
...
+        socket.emit("updateNote", this.as_dict());
+    }
+
+    show(): void {
+        this.dialog.dialog({title: this.name});
+        this.dialog.children().eq(0).val(this.text);
+        this.dialog.dialog("open");
+    }
+
+    as_dict(): Dictionary<string> {
```
Kruptein: I would suggest changing this to `asDict` as typescript uses camel casing and not snake casing like python.

ts_src/manager.ts
AntiSC2: While that solution is more elegant is doesn't work in this case because of this line: `gameManager.notes.get(uuid).show();`

Typescript throws an error here because orderedMap.get expects a string as an argument and Typescript thinks that uuid can be a string or undefined.

Kruptein: Ah yes because it's reassigned to uuid itself which is `string || undefined` my bad, the shape constructor assigns to this.uuid which is solely `string`.  I suppose you could use `new Note(name, text, uuid || uuidv4())`, but you may leave it as you have it now as well

AntiSC2:
@Kruptein, have made all the changes now

Kruptein:
A last small nitpick, you can change:
```typescript
uuid = uuid || uuidv4();
const id = <string>uuid;
```
into 
```typescript
const id: string = uuid || uuidv4();
```

And the `: string` part is optional, because the new variable is automatically typed as `string`.

Additionally, given that you increased the save version, I would also like to have an upgrade script for older versions.  I've quickly created something which I believe should be enough [here](https://gist.github.com/Kruptein/3c9b24a9fe296159790604f0cdb1ae18), but I've not tested it yet.  This file should be added to scripts/convert/

AntiSC2:
@Kruptein, fixed the last small nitpick and added the convertion script. Did only a simple and quick test of it but it seemed to work.

PR #73, opened 2018-09-27
Add dockerization for easy deployment

schemen:
Also, I've changed the way configs are loaded. Also added is that we can change the savefile location.


Kruptein:
Hey, can you do the pr on the dev branch instead? I follow gitflow which uses the master is a release branch.

Also ignore the appveyor build, i have yet to set it up

schemen:
Hey, I just changed the into branch to dev :)  you can use this PR still :)

Kruptein:
Oh, silly me, i'll look at the pr more in depth tonight

schemen:
Cheers! Let me know if you want anything changed. This branch is currently deployed to my cluster at https://planarally.dndbox.com

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+WORKDIR /usr/src/app
+
+VOLUME /usr/src/app/data
+VOLUME /usr/src/app/static/assets
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+
+COPY requirements.txt ./
+RUN apt-get update && apt-get install dumb-init && \
```
FrederikGoovaerts: I think it's a good idea to do `rm -rf /var/lib/apt/lists/*` after the install completed. The docker container won't need the apt list anymore and this takes up a decent amount of space.

Dockerfiles/server_config.cfg
```diff
@@ -0,0 +1,9 @@
+[Webserver]
```
FrederikGoovaerts: Why is this file duplicated now? If I interpret correctly, you copy everything in the Planarally folder, including the cfg file there, then copy this one over the original one. Am I missing something?

PlanarAlly/planarserver.py
```diff
@@ -871,9 +875,7 @@ def shape_wrap(player, shape):
 app.on_shutdown.append(on_shutdown)
 
 if __name__ == '__main__':
-    app['background_save'] = sio.start_background_task(save_all)
-    cfg = configparser.ConfigParser()
-    cfg.read("server_config.cfg")
+    sio.start_background_task(save_all)
```
Kruptein: You want to keep the app['background_save'] assignment, as it is used on line 84 to cancel the task.

schemen: Done that!

Dockerfile
schemen: Good Idea, done that.

Dockerfiles/server_config.cfg
schemen: I'm only copying the content of the folder PlanarAlly/, not the entire thing.

PlanarAlly/planarserver.py
Kruptein: Now you start the background save twice ;)

schemen: I have updated it, do you mean like that?

Kruptein: indeed

schemen:
Ok, so far all remarks are resolved. Anything else I could improve :) ?

Issue #75, opened 2018-10-01
Cannot move inititive window in Firefox


W8BSD:
While the initiative window can be moved easily in Chrome, in Firefox it doesn't seem to be movable.

Running Firefox 61.0.2 (64-bit) on FreeBSD.

W8BSD:
Same issue with the Edit Asset window.

Kruptein:
The drag feature for modals and dialogs indeed seems to be broken on
Firefox on all platforms

PR #76, opened 2018-10-01
Proof-of-concept fixing Firefox modal drag-drop

W8BSD:
Hack to fix https://github.com/Kruptein/PlanarAlly/issues/75

I don't have node set up on my box, but when I made the changes directly to the js file, it appears to have worked fine.  It also seems like the if() isn't actually needed since the hack works in Chrome too, but I was being paranoid about the change.

Only issue is that in Firefox, the dragged image is offset down and to the right a bit from where the window will actually end up.

Two main changes here... the simple one is that event.dataTransfer.setData() is required for the drag to be shown on Firefox.  If use "text/plain" or "Text", it navigates to a different page on drop... using "Hack" appears to prevent this for unknown reasons.

The big one however is that Firefox doesn't populate clientX/Y or pageX/Y for drop events.  The bug is here: https://bugzilla.mozilla.org/show_bug.cgi?id=505521 and appears that it is unlikely to be fixed.  This hack uses screenX/Y which does appear to be supported in Firefox.

**This patch has not even been build tested**


Kruptein:
I'll look at your changes when I'm on my pc, I've fixed the dragging on Firefox partially on a separate branch, i figure i had to set data to make it draggable. It does however open a blank tab on drop so your suggested fix is greatly appreciated.

Can you change your pr to the dev branch instead of the master branch btw?

W8BSD:
Ok, that was clearly the wrong way to change the base branch. :(

Kruptein:
I adjusted something to the changelog hence the conflict, can you however immediately take the opportunity to add your fix to the changelog under unreleased/fixed ?

Kruptein:
Also something weird seems to have happened during the base change with the readme, just copy the current readme from the dev branch ;)

PR #79, opened 2018-10-06
Add asset rotation support

W8BSD:
This allows for using tile sets for building maps.  Currently can only
rotate 90 degrees in either direction.

Make sure to do your PR on the `dev` branch and NOT on the `master` branch.  This repo uses gitflow which only uses the master branch as a release branch.

Make sure to update the `CHANGELOG`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
     - The server config now has an option to specify a different save file name and/or location.
 - Dockerfile [contributed by Schemen]
     - A dockerfile is now present to support deployment in docker containers
+- Allow rotating assets 90 degrees at a time
 
 ### Fixed
 - Unable to drag modal dialogs in Firefox
--- a/ts_src/game/api_types.ts
+++ b/ts_src/game/api_types.ts
@@ -117,6 +117,7 @@ export interface ServerText extends ServerShape {
 }
 export interface ServerAsset extends ServerRect {
     src: string;
+    angle: number;
 }
 
 export interface Note {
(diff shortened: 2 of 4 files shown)


W8BSD:
Hah, got it.  Much nicer.

Kruptein:
I intend to add rotation at any angle somewhere in the future, but for that to work I'll first have to adapt the lighting algorithm to work with non axis-bound bounding boxes.

So for now your addition will work fine, I took a quick look at the code and it look clean, I'm gonna test it later today.

Can you update the changelog ? :)

Kruptein:
Some remarks:

* The type field is protected on Shape, which results in typescript errors in your shapecontext.vue changes.  By changing the type === 'asset' check to an instanceof Asset check, these errors are gone and also immediately make sure that 'angle' actually exists on the shape object.
* Your rotations are currently not synced to the server, which means that other players won't see your changes and a reload will also reset the rotation.  You'll want to use  `socket.emit("updateShape", { shape: shape.asDict(), true });` the second argument tells the server whether the shape change requires a redraw which it does in this case

W8BSD:
Thanks, I didn't see the error with === 'asset'... and I must have tested other screens when I was still using the remove/add method to force the draw.

I can't figure out how to use instanceof Asset in the v-if bit though.

Kruptein:
[this issue](https://github.com/vuejs/vue/issues/7708) has some info on how you could fix it, but I guess it's fine for now to just use the === 'asset' in the v-if check

Kruptein:
You should probably also only allow the rotation on a shape if you own the shape.  You can check this with `shape.ownedBy()`.  The function accepts an optional username, but defaults to the current user.

I notice that this is not checked for the move to front/back functions but that should be fixed as well in the future

Kruptein:
I'm gonna close this PR for now, given the lack of updates and the fact that the dev branch has had substantial changes, so the changes in this PR need to be redone anyway

PR #86, opened 2018-10-10
Rename all event names to new format

### LINKED ISSUE #82: Clean up event names
Socket.io eventnames are currently inconsistent in naming scheme.

I'm fond of using something like the initiative events have, but this should be continued in the other event names.

AntiSC2:
Eventnames follow a standard now:
- Every name starts with uppercase
- Name of object that is sent is first
- What action is performed comes after name
- Actions start with uppercase and is seperated by .
- When client recieves information it is by a Get action
- When server recieves information it is by a Set action

Please check that the new names are ok. Before merging I want to squash the commits.

Fixes #82.


AntiSC2:
Also important, testing. This pull request can break a lot of things. Have of course tested myself but I haven't been able to cover everything.

Kruptein:
At first glance most names are as I want them.

I'm only not really sold on the 'Get' names for the client.  It seems counter intuitive to me to call for example `GridSize.Get` when it clearly sets the gridsize.

Also there are some places where you used set instead of get e.g. Initiative.Set on line 445 of the server.

So good work, but I would prefer to just use the .Set directives unless you have a strong argumentation against it.

AntiSC2:
It was just to seperate when sending and receiving information but I agree that it doesn't make much sense in some situations. Will rename and change a bit.

AntiSC2:
Closing this as I'm for some reason having trouble pushing to this branch.

PR #88, opened 2018-10-13
Make the tools start with an uppercase

AntiSC2:
- Also remove link border when clicking in the radial menu
- Fixed a missing space on the remove button in the note modal

The reason for this change is because I think it looks much better when it starts with an uppercase


Kruptein:
Update the changleog, rebuild the scripts to fix the template conflicts and it's good to go.

AntiSC2:
Will do

AntiSC2:
Done

PR #89, opened 2018-10-25
Clarify Installation instructions

AchielC:
Make sure to do your PR on the `dev` branch and NOT on the `master` branch.  This repo uses gitflow which only uses the master branch as a release branch.

Make sure to update the `CHANGELOG`.


Kruptein:
I would like to make sure that it is obvious that only the DM has to install the software.  I feel like currently a player will look at this and think he'll have to install it as well.

PR #107, opened 2018-11-12
Fix & extend docker image

schemen:


schemen:
With that, https://planarally.dndbox.com is on version 0.11.6 :)

Kruptein:
Nice, just out of curiosity what's curl used for ? I just see it in the apt-get install list but nowhere actually used.

(I know what curl does, just wondering where it's used specifically)

schemen:
I use it for the healthcheck to see if the container/programm is running healthily.

I noticed that sometimes after long use by someone, planarally crashes for some reason without much indication. The container kept running in a broken state but the website didn't deliver anything.

I created a healthcheck that basically runs a curl every few seconds to see if it still responds - if not, it restarts the container after a timeout.

Kruptein:
hm ok good to know, if you ever find out what causes it let me know.

I have the feeling that the underlying webserver sometimes has some async stuff pending.

PR #128, opened 2019-02-02
Set up CI with Azure Pipelines

Kruptein:


Kruptein:
CircleCI has proven to just work, whereas I've not been able to get azure to work.

Issue #150, opened 2019-07-31
Sessions with a slash in their name do not work


CalebKoch:
Sessions with a '/' in their name do not work correctly. You can create the session, but when trying to load it, I just see a blank background. 

Seems very related to #20

PR #151, opened 2019-07-31
Feature/encode slash

CalebKoch:
Encode the values of creator and game whenever they are set,


CalebKoch:
Addresses https://github.com/Kruptein/PlanarAlly/issues/150

Kruptein:
I'm gonna run it later to verify some behaviour, before merging, but otherwise it looks pretty good!

Thanks for the work :).

PR #159, opened 2019-09-15
Added ping tool

Daniferrito:
It is literally a reskined ruler tool


Daniferrito:
The failing test seems to be from a misconfigured script, as npm is not recognizing the lint command.

According to my local lint, there does not seem to be any lint errors

Kruptein:
Hey, the linting failing is indeed a normal process for all current PRs on dev, because I swapped out tslint for eslint, but this is still in a feature branch and hasn't been merged yet.  (The CI addition also coincides with that branch hence the current breakage on 'normal' PRs).

This is something I have to fix, so no worries.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Changed
```
Kruptein:
```suggestion
### Added
```

Kruptein:
I've tested it in the meanwhile, works fine. This is good to go, the only thing I could suggest is maybe having the radius of the circle change to draw attention, I can imagine the circle being less obvious in a crowded map. (although just moving it around also works albeit slightly less elegant).

If you would go that route I would strongly suggest to use a invisible fill, and a border colour instead of just the circle becoming larger.

If you feel this is less ideal, I'm fine with merging it as is.

Daniferrito:
> having the radius of the circle change (during mousemove) to draw attention
I'm not sure what you mean. Something like make the circle 2x the size when moving, then shrink when it has stopped?

> If you would go that route I would strongly suggest to use an invisible fill, and a border colour instead of just the circle becoming larger.
I feel that would only make it less visible. While it allows to see what is under the circle, you can also get that by stopping to click.

Two extra things I want to add to this:
Right mouse to spawn a roll20 kind of ping (maybe going inwards instead of outwards) (that is, a circumference that starts quite big (maybe 2x tile radius), and becomes smaller over 1 second, ending up as a point in the clicked position)
An indicator for players that are looking other places in the border of the screen, towards the ping, so that the ping is always noticed.

This changes are more complex, as they require the clients to know about pings themselves, so they can:
* Animate the ping themselves (all frame's updates coud be pushed from the originator through the net, but this seems way too wastefull)
* Draw the arrow towards the ping (each client has to calculate its own arrow direction, and whether to even show the arrow)

As this requires more complicated changes, I would implement them in a future update. I will implement shortly the bigger circle when moving.

Daniferrito:
There it is. I dont like how it looks. Maybe an animation to grow/shrink instead of the sharp change might solve it, but we are back at the problem of animating through the net or adding capabilities to the client to run remote animations.

I would suggest to revert the last commit so the circle does not rescale and merge.

Kruptein:
I've not run it myself, but if you think it's ugly, then I'm fine with reverting and revisiting this when some kind of animation support is added.

Another random idea: 2 circles.  one that is an actual filled circle like you have now and another one with a bigger radius that is filled invisible but has a border ?

Daniferrito:
Here is the ping tool as it is now. It does look better, but I'm not totally happy with it. Not much else I can think of doing to it, and it is indeed functional, so i vote for leaving it as it for now. 
I left the code for the movement stop detection. It does send an extra unnecessary event, but it should be negligible.

I tried to improve the throttle code to prevent the checking of the "has been initialized before" each time, and I promise it DID work at some time, but after returning to this branch it just refused to work any more. The same kind of code is still working in the ping tool, so not sure what happened. You can see the commit history if you are curious.

How the ping tool looks like:
Screenshot_2019-09-22_14-58-42

Kruptein:
Hey, can you explain what you're trying to achieve with the debounce and the mouseStop function?

I've recently added a lodash/throttle on all incoming mouse move events in game.vue, to prevent too many events from firing.

And even with the debounce the shapes are still updating on mouse move anyway ?

You can also get rid of the radius reassignments now that it's no longer changing.

Daniferrito:
> Hey, can you explain what you're trying to achieve with the debounce and the mouseStop function?

I'll explain this with much more detail than is probably needed, just in case, so bear with me (if the question was "Why is the debounce still there when it does nothing now?", the answer is "For when the animations are added", and you should be free to skip the rest of the explanation):

Ok, so we are interested in the following events:
When the mouse is pressed: To make the dot appear
When the mouse is released: To make the dot dissapear
When the mouse is moved: To make the dot move with the mouse
When the mouse starts moving: To make the radius bigger
When the mouse stops moving: To reduce the radius to the normal size.

The first three are easy, as they are literally events sent by the browser (and the ones that existed originally).
For the "starts moving", we could store a variable "is moving", and every time the mouse is moved, if "is moving" was false, run the code asociated with "starts moving" and set "is moving to true". Then set "is moving" to false in the "stops moving" event. However, as the operation in the "starts moving" is idempotent, we can just run it every time.
For the "stops moving", it is harder. If we had code that ran each frame, we could check if the "is moved" method was called this frame, and run the "stops moving" in a similar way than the proposed in the last point. But we have two problems. We dont have code that runs each frame synchronously with the "is moved" function, and we dont want a single frame of non-movement to run the "stops movement" code. We would prefer a bit of leeway, so only the "is moved" not running for half a second should run the "stops moving", but is the user keeps moving the mouse, it should not trigger. That is literally what the debounce does.

Intermission: Difference between throttle and debounce:
I'm not sure where throttle and debounce come from originally, but I learnt of them from ReactiveX (usually Rx or RxJs). 
Rx defines observables, or streams, where events come from one side, some operations are performed with the events, and they come out the other way. This operations can be many things, like the usual map (transform the events) or filter (only let some events through), but the interesting ones for this are the time-dependant operations. where the exact time each event (and the surrounding events) go through the stream matters.
Throttle and debounce are such operations. They also have in common that they recieve one duration, and the events in the outputs will never be less than that duration between them.
How exactly they do that is the difference between throttle and debounce. 
Throttle will let through the first event, and then discard any event until the duration has passed, at which time it resets.
Debounce will start a timer for the duration when the first event arrives. When that timer ends, it will emit the event through the output. However, if another event arrivesbefore the timer from the first event ends, the first event will be discarded, and a new timer will be started.
They have advantages and drawbacks (althrough sometimes those "drawbacks" are the intended behaviour). Throttle will have no initial delay, while debounce is guaranteed to have a delay of the specified duration. For example (with a duration of 500ms), if you throttle mouse movement, it will be choppy, but as soon as the mouse if moved, the shape will move. If you debounce it, no matter what the user does, it will have a "lag" of 500ms. Even more, if the user keeps moving the mouse, the debounce will never receve an event in the output. On the other hand, once the mouse stops moving, with throttling you will have a position as the last event, but it will not necessarly be the last position of the mouse (the throttle will have probably dropped that one). However, with debounce, the last position (once it arrives), will really be the last position of the mouse.
In general, they have different applications. Usually, a searchbox is debounced, to prevent each letter added to send a new search to the server, but you need the final inputed thing to be searched. Mouse clicks, on the other hand, are a better fit for debounce. Clicking needs to be instant, so the "lag" is not acceptable, but if a user clicks on a shape 5 times in a second, the code to select the shape only needs to be run once. Mouse movement, also, cannot be debounced, as it could happen that the user keeps moving the mouse, and you are interested in the intermediate events to display feedback to the user.
For other set of examples, sometimes the event themselves are not useful, but only that they happened. In a server, it is useful to debounce events from the clients (on top of processing them). Then, if any event comes from the output of that stream, the server knows that that client has probably disconnected, as it has stopped sending events. On the other hand, you could generate "stay alive" events from the client from all input events from the browser. However, as you do not want to send multiple events per second to the server, you should throttle them so the server is not overwhelmed (and the exact event that produced each "stay alive" is not useful at all). If you however wanted a "afk" detected from the client, you should go back to debounce.
For a usefull page on observables, and the exact behaviour of the operations on them, see https://rxmarbles.com/ debounce and throttle are called debounceTime and throttleTime there.

The lodash versions of throttle and debounce are a bit more limited on what they can do, but for the current necessities seem suficient (if lodash was being used anyway).

> I've recently added a lodash/throttle on all incoming mouse move events in game.vue, to prevent too many events from firing.
> 
> And even with the debounce the shapes are still updating on mouse move anyway ?

As I said, debounce is probably not a good option for mouse movement. As we are interested in the last event it might be interesting to combine throttle and debounce (in paralel, so either function outputing triggers an event), so we get both intermediate feedback and the final position. I believe this would be a similar solution to just using the trailing option in lodash (but I do not know of a similar option in RxJs)
For the problem you report, I would need to look into it to see what is happening. Could you tell me what is happening, what would you expect (or want) to happen, and the related code to the issue you are trying to solve? I expect this to be the throttle in mousemove in game.vue, where tools.mousemove is being called more than once per 15ms period?

> You can also get rid of the radius reassignments now that it's no longer changing.

I know, but the assignement is basically free (setting a variable, and the update to the server needs to be done anyway as the circle moved), and it serves as a reminder to add the animations later. I can remove it anyway if you would prefer.

Kruptein:
Thanks for the detailed write-up.  Honestly I would just remove the stuff that is not in actual use for now, as it will only add an extra layer of potential confusion for others or even myself.  I also think that for animations we might look for another approach entirely anyway.  You mention the lack of some code that runs each frame.  This actually does exist.  There is a core game loop in `layers/manager.ts` `drawLoop`.  This code goes over all layers and requests a new draw on each frame.  Obviously you don't want to constantly redraw every static thing, so each layer has a `valid` field to indicate whether it is still valid (aka didn't change) since the last draw.

At some point I would invest in adding some sort of animate logic to shapes in general, which would make use of this game loop.  For now there are no concrete implementation plans, but I suspect it would be more interesting to add it in a generic way for all shapes instead of doing debounce logic per shape. (unless you disagree).

Daniferrito:
> Honestly I would just remove the stuff that is not in actual use for now, as it will only add an extra layer of potential confusion

Fair enough, i will remove it.

> but I suspect it would be more interesting to add it in a generic way for all shapes instead of doing debounce logic per shape

Oh, no, the debounce is only to trigger the shrink animation (both localy and sending an event to the server). Running the animation would need to be its own thing.

PR #160, opened 2019-09-15
Added dropdown for the selection of length unit

### LINKED ISSUE #94: Add toggle between ft and m
Option to choose metric system instead of imperial system

Daniferrito:
Also changed all references to ft to use this selection

Closes #94


Daniferrito:
We might want to split this to its own selection, to allow for any kind of units other than ft and m.

Especially thinking on the suggestions from #164:
"This allows for some neat features, like an overworld map, where distance can be measured in kilometers,"
Instead of adding every single possibility, just have the two default ones and a third option to freefil whatever the DM wants.

I'll add that option in its own row.

Daniferrito:
As I said in the commit message, i wanted to use a dropbown with optional free input, but could not make one that was both nice and intuitive. Many options at https://stackoverflow.com/questions/18309059/drop-down-menu-text-field-in-one

The change to fully free options is to allow for other options we have not tough yet, like the previously mentioned overword map in kilometers or miles, or maybe space tactical battles, or space strategy maps ranging whole galaxies, wich might have units of light-years, parsecs, kiloparsecs...

And if someone wants to measure their distance in [smoots](https://en.wikipedia.org/wiki/Smoot), as explained [https://www.youtube.com/watch?v=-scs_yF59YE](url), they should be able to!
Smoots

Kruptein:
So I ran the code, client side it seems fine, there is however an important part missing being the server side :).

You'll need to update the database model to have this new element available.

To do this:
(1) Find the server/models/campaign.py file and add an extra field to the Location class.
(2) Modify server/save.py
    (a) Update SAVE_VERSION by one increment
    (b) Add an extra clause to the big if thingy for upgrading from version 16 of the save format

The extra clause should be very similar to the last save format change as it's just a simple column addition and nothing too complex.

If you want to, I can also create a gist with the code you need, but feel free to give it a try yourself as well.

Rewriting that file to not be such an eye sore is also on my list of things to do :p  It should ideally also not use a single save version as that can become a mess with multiple PRs changing save format. but nothing for you to worry about now.

Daniferrito:
I did not know i would need to add an extra field to the database. I supose it works because it is being kept in memory, but will reset when the server resets.

Kruptein:
Hey, no problem at all. I'm already surprised with how well you were able to traverse the code base so far, without any documentation

Daniferrito:
Done. Let me know if it looks ok.

We might want to add the instructions from your comment into the documentation somewhere. It would not be complete documentation, but incomplete documentation is better than no documentation, and the cost of copy-pasting and lightly editing it is quite minimal.

I have experience navigating and analyzins strange code for various reasons. However, I have to say that PlanarAlly is quite nicely written and structured so it was quite easy to follow. At least the frontend and the http side of the server. The server-side socket part I find much harder to navigate and get a feel for how it interacts with the client.

server/save.py
```diff
@@ -223,6 +223,17 @@ def upgrade(version):
             )
         db.foreign_keys = True
         Constants.update(save_version=Constants.save_version + 1).execute()
+    elif version == 16:
+        from peewee import TextField
+        migrator = SqliteMigrator(db)
+        db.foreign_keys = False
+        with db.atomic():
+            migrate(
+                migrator.add_column("location_user_option", "unit_size_unit",
```
Kruptein:
```suggestion
                migrator.add_column("location", "unit_size_unit",
```
the LocationUserOption table is another table than the Location table.  The former are some client specific settings like the current view position in the location, whereas the latter has the concrete settings that apply to everyone.

Issue #129, opened 2019-02-05
In-Person use improvements


schemen:
Heya!

My main usage for PlanarAlly is for in-person use (A flat screen on the table). Any chance we could add/improve some features?
I had some ideas in my mind of the likes:

* Make it possible to scale the player view from the dm screen (so I can get the correct size for minies)
* Remove all UI elements on the player view
* Simpler "Move Players here" method, currently only possible with the select tool and right click. Maybe have it on right click all the time on all layers

This could extend to a In-Person mode which removes all the dynamic lighting stuff to declutter UI elements.

Cheers!

Kruptein:
Hey,

Could you clarify some things?

What exactly do you mean with scale a player view as the dm? Do you mean
controlling the zoom level of that player from the dm side while still
having a separate dm zoom level, or simply mimicking the dm zoom?

Could you elaborate in which way the lighting clutters the ui in your
opinion?

Thanks for using PA and providing feedback, I'll see what I can do :)

schemen:
Yeah of course!

> What exactly do you mean with scale a player view as the dm? Do you mean controlling the zoom level of that player from the dm side while still having a separate dm zoom level, or simply mimicking the dm zoom?
Hm, both sound good. First one could be achieved by having two sliders - secondly could be achieved easier I suppose. But yes, I mean the zoom level so the minies fit the screen. The "Bring Players here" is already available and great, now something like that for zoom as well :) or maybe even a static setting per map (like FOW Opacity level - as screen size doesn't really change and thus the zoom doesn't as well - I might even like this option the best, lol)

> Could you elaborate in which way the lighting clutters the ui in your opinion?
I just thought it would make the UI more lightweight, I might be off though!


Btw, planarally on dndbox is running very stable, looking at the logs & resource stats I am very happy with it! There is actually quiet a few people using it fairly regular :D

Kruptein:
Hm, I'll have to think about it to make sure I add it in such a way that it does not hamper any other users.

That being said, disabling the UI is definitely easy to add.

Nice to hear that it's running smooth!

AJAnderson:
I second being able to disable the UI, particularly on the player screen.

Kruptein:
Toggling of UI can now be done on the dev branch with `ctrl+u`

dthv:
As far as I see it, there is only one part of this issue that can be considered still open:

>     * Make it possible to scale the player view from the dm screen (so I can get the correct size for minies)

[...]

Issue #97, opened 2018-11-08
Move token to other location


Purzelkater:
It would be nice, if tokens could be moved to other locations (maps/pages). This would it make easier to handle with player tokens. At to moment, every token has to be inserted on every location and all tracker and aura settings manually edited.

Kruptein:
Interesting idea, I'll add it to my list of todo's :)

Daniferrito:
It could e an option in the right click menu, just as moving tokens to another layer. I will probably take a look at this

Kruptein:
Either that and/or something with dragging a token on a location.

I've plans to change the way the location bar at the top functions (as tbh it currently is rather ugly).

But by all means go ahead and add it to the context menu :)

dthv:
Is a drag/drop option with the new top-bar planned?

Kruptein:
there isn't but it shouldn't be too hard to implement.

PR #161, opened 2019-09-15
Implemented #129 plus more

Daniferrito:
Move players here also sets the zoom
The default right click menu is now accessible in all tools (plus renamed the class to defaultcontext)
Right clicking with over a non-selected shape now selects the shape and shows its options as if it was selected beforehand
Right clicking over a selected shape no longer deselects the rest of the shapes, and allows executing the actions on all shapes at the same time
Right clicking a selected shape now only shows the show properties option if it is the only shape in the selection


Daniferrito:
Implemented #97, plus added ctrl and shift multiselection (both keys do the same)

client/src/game/ui/selection/shapecontext.vue
```diff
@@ -19,85 +19,139 @@
...
+        this.close();
+    }
+    setLocation(newLocation: string) {
+        if (this.shapes === null) return;
+        const layer = layerManager.getLayer()!;
+        layer.selection = this.shapes;
+        cutShapes();
+        // Request change to other location
+        socket.emit("Location.Change", newLocation);
+        socket.once("Location.Set", (_data: Partial<ServerLocation>) => {
```
Kruptein: Do you know if there is any guarantee that this is always ran after the main listener in events.ts ?  Depending on which one fires first, the lighting is updated with the new shapes or without.

client/src/game/ui/selection/shapecontext.vue
```diff
@@ -19,85 +19,139 @@
...
+            // Paste when location changes
+            // TODO: Shapes are pasted to map layer independently of where they come from. Fix this
```
Kruptein: You could alter the pasteShapes signature to 
```typescript
export function pasteShapes(offsetPosition: boolean = true, targetLayer?: string): Shape[] {
    const layer = layerManager.getLayer(targetLayer);
```
which will default to the current layer for normal ctrl+v, and use a specific layer in this part. When the target layer however does not exist, no shapes will actually be pasted as the function will early return.

Another approach is to simply call `selectLayer(layer.name)` in `layers/manager.ts` before pasting.  If the layer does not exist, selectLayer will basically act as a no-op, and the shapes will thus be pasted on the current active layer of that location (which should be the last active layer for the current user on that location).

Your current approach with calling setLayer after pasting them, gives some small communication overhead, given that you first paste all the shapes to then all move them again.  but otherwise is fairly similar to the selectLayer approach.

In practice currently, every location should have the same set of layers, so the aforementioned problems are not really relevant.  I do plan however on adding more flexibility in the layer system in the future, which could result in different sets of layers to exist across locations, hence the thought exercise. At that point either a dialog needs to appear to choose a target layer, or if possible a target suited for tokens is automatically selected.  But hey that's far away.

client/src/game/ui/tools/select.vue
```diff
@@ -190,7 +192,11 @@ export default class SelectTool extends Tool {
         const layer = layerManager.getLayer()!;
 
         if (this.mode === SelectOperations.GroupSelect) {
-            layer.clearSelection();
+            if (keyboardState.shiftPressed || keyboardState.ctrlPressed) {
+                // If either control or shift are pressed, do not remove selection
```
Kruptein: If you draw a new box in shiftmode that also contains a shape that already was in a selection, it now is included twice in the selection.  This needs to be handled in line 211!

client/src/game/ui/tools/select.vue
```diff
@@ -111,7 +109,11 @@ export default class SelectTool extends Tool {
             this.selectionHelper.w = 0;
             this.selectionHelper.h = 0;
 
-            layer.selection = [this.selectionHelper];
+            if (keyboardState.shiftPressed || keyboardState.ctrlPressed) {
+                // If either control or shift are pressed, do not remove selection
```
Kruptein: In this case and the above, the selection helper should respectively be increased or added to have consistent behaviour.

Kruptein:
With further playing around with some of the new things I noticed two other things:

(1) Location changing is broken in some cases, this is introduced by me in the big pr, so I'm gonna make a small pr that fixes this.
(2) When you move objects from one location to another, they are placed at their original (x,y) position.  The camera however is focused on the place where it last was when visiting that location.  Which often will not coincide.  To solve this you'll have to either:
(a) Move the shapes to the current view point in the new location
(b) Move the camera to the location of the new shapes

The latter is easier, but does add the risk that you're suddenly ported to somewhere far away from the actual map and with no current good UI to find your map back, this can be rather annoying.

client/src/game/ui/selection/shapecontext.vue
Daniferrito: Not sure. It might be guaranteed that recieving events are recieved in order of subscription. I think the correct solution would be to first emit a "paste to location" event, then change locations. However, i'm not sure if emiting events guarantees that the server recieves them in the same order, or that the first event will be totally processe before running the second one.

client/src/game/ui/selection/shapecontext.vue
Daniferrito: Calling selectLayer before pasting changes the selected layer in the destination location. This should not be a problem, though.

I will change the pasteShapes signature as suggested, i like it.

client/src/game/ui/tools/select.vue
Daniferrito: I did not notice this. I believe this would be especially problematic when pasting shapes. I will add a check to not add shapes to the selection multiple times.

client/src/game/ui/tools/select.vue
Daniferrito: I do not understaund what you mean.

Daniferrito:
> (a) Move the shapes to the current view point in the new location

I would prefer this, and it will probably be what I will probably implement.

client/src/game/events/keyboard.ts
```diff
@@ -1,42 +1,44 @@
...
+    keyboardState.shiftPressed = event.getModifierState("Shift");
+    keyboardState.ctrlPressed = event.getModifierState("Control");
+    keyboardState.altPressed = event.getModifierState("Alt");
+}
 
 export function onKeyUp(event: KeyboardEvent): void {
-    if (!(event.target instanceof HTMLInputElement || event.target instanceof HTMLTextAreaElement)) {
+    updateKeyboardState(event);
+    if (event.target instanceof HTMLInputElement || event.target instanceof HTMLTextAreaElement) {
+        noop;
```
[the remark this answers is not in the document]
Kruptein: You even used it yourself unknowingly! :D 

```
if (keyboardState.shiftPressed || keyboardState.ctrlPressed) {
    // If either control or shift are pressed, do not remove selection
} else {```

File: client/src/game/ui/selection/shapecontext.vue [hunk 6]
```diff
@@ -19,85 +19,139 @@
...
-        if (this.shape === null) return;
-        this.shape.moveLayer(newLayer, true);
+        if (this.shapes === null) return;
+        this.shapes.forEach(shape => shape.moveLayer(newLayer, true));
+        this.close();
+    }
+    setLocation(newLocation: string) {
+        if (this.shapes === null) return;
+        const layer = layerManager.getLayer()!;
+        layer.selection = this.shapes;
```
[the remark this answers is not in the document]
Kruptein: Right, I was wondering why you reasign the layer selection, I was under the impression this would already equal this.shapes at this instance, I might be wrong though.

client/src/game/ui/selection/shapecontext.vue
Kruptein: It's plausible that they are processed in subscription order yes.

If I remember correctly the messages are guaranteed to arrive in the correct order. (at least in the case where you send multiple messages to others; if multiple people were sending stuff to you then the latency would be the deciding factor)

client/src/game/ui/selection/shapecontext.vue
Kruptein: Isn't this what you would expect though ? Or would you assume to keep the original selected layer selected after the paste?

Location A: active layer 'tokens'; Location B: active layer: 'fow'
Move selection of tokens to location B, they will be properly pasted on location B's tokens layer.

What is the expected active layer after the paste happens in location B? tokens or fow?  I was under the impression it would become tokens as chances are high that you'll want to move the tokens you just pasted around.

client/src/game/ui/tools/select.vue
Kruptein: Before starting the new selection with shift, a selection box is already present, after dragging the new box, the old one is removed but there are still elements from that selectionbox selected, but possibly outside of the new box.

This feels kinda strange, so I would suggest to make the original selectionbox bigger to fit all previous as well as all new shapes.  Or we just get rid of the selectionbox all together as you alluded to earlier.

client/src/game/ui/selection/shapecontext.vue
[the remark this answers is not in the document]
Daniferrito: Just in case. Probably the correct solution would be to remove this.shapes and just use the layer.selection array and not have the two diferent copies.

client/src/game/ui/selection/shapecontext.vue
Daniferrito: I assume the messages would be recieved the right order. I mean if the first message would be completelly processed before the second message starts to be processed.

client/src/game/ui/selection/shapecontext.vue
Daniferrito: You are right. Probably even if the paste method gets refactored or if the server does the pasting before the layer is even changed, we might want to change the active layer anyway.

client/src/game/ui/selection/shapecontext.vue
Daniferrito: In any case, we could just send a message back from the server with "onPasteFinished" or similar. Now we have the problem of what happens if the server finishes the paste before the client sets the callback. Probably just seting the callback before doing the paste would work anyway.

client/src/game/ui/tools/select.vue
Daniferrito: I do not like making the previous selectionBox bigger, as that would cover shapes that the user does not want to select, and that would be confusing. Just removing the selectionbox would be cleaner IMO.

Kruptein: I can live with that, but that's for another PR, so this is OK.

client/src/game/ui/selection/shapecontext.vue
Kruptein: I just checked the node docs and they are in the order they are registered. so the .once() will in this case always be called after the main .on() in events.ts

Daniferrito:
Ok, mostly done. Only one issue left (but it is a big one):
> When you move objects from one location to another, they are placed at their original (x,y) position
The correct solution (IMO) would be to copy them relative to the new location. That is, if the camera is centered in 500,0 in the first location, and in 0,200 in the second location, all the objects get their position offset by -500,200 before pasting. However I've had no luck getting this information from the available information.

What I've got:
Subscribe once to Location.Set, then in the subscription subscribe to Client.Options.Set. Once that arrives, all the necessary data should have arrived. As javascript is single-threaded, setting the subscription is guaranteed to finish before the event is finished processing, and all the subscription from one event should finish before the events from the second event trigger. The server always calls Location.Set before Client.Options.Set (from server/api/socker/location.py load_location())
Change pasteShapes to recieve an optional offset instead of a boolean. If this offset is not set, it defaults to 10,10.

client/src/game/ui/selection/shapecontext.vue
```diff
@@ -19,85 +19,133 @@
...
+    }
+    setLocation(newLocation: string) {
+        const layer = layerManager.getLayer()!;
+        cutShapes();
+        // Request change to other location
+        socket.emit("Location.Change", newLocation);
+        socket.once("Location.Set", (_data: Partial<ServerLocation>) => {
+            socket.once("Client.Options.Set", (_options: ServerClient) => {
+                // TODO: Shapes are pasted to their original position instead of the position the camera is now centered
+                layer.selection = pasteShapes(new Vector(0, 0), layer.name);
```
Kruptein:
```suggestion
                layer.selection = pasteShapes(undefined, layer.name);
```
I understand the reason why you chose new Vector(0, 0) here, but it's an unnecessary allocation, that if left as undefined, will jump straight over the if (offset) line with the same end-effect.

client/src/game/ui/tools/select.vue
```diff
@@ -202,14 +207,12 @@ export default class SelectTool extends Tool {
                     this.selectionHelper!.refPoint.y <= bbox.botLeft.y &&
                     this.selectionHelper!.refPoint.y + this.selectionHelper!.h >= bbox.topLeft.y
                 ) {
-                    layer.selection.push(shape);
+                    if (layer.selection.find(it => it == shape) === undefined) {
```
Kruptein:
```suggestion
                    if (layer.selection.find(it => it === shape) === undefined) {
```

client/src/game/ui/tools/select.vue
```diff
@@ -202,14 +207,12 @@ export default class SelectTool extends Tool {
...
+                        layer.selection.push(shape);
+                    }
                 }
             });
-
-            // Push the selection helper as the last element of the selection
-            // This makes sure that it will be the first one to be hit in the hit detection onMouseDown
-            if (layer.selection.length > 0) layer.selection.push(this.selectionHelper);
-
+            layer.selection = layer.selection.filter(it => it != this.selectionHelper);
```
Kruptein:
```suggestion
            layer.selection = layer.selection.filter(it => it !== this.selectionHelper);
```

Daniferrito:
And that should be everything. It would be easier to move shapes relative to the top-left of the screen, as it avoids having to calculate where the center is, but that changes the new position a bit more.

The other option would be to move relative to the mouse position. We do not have the mouse position after the screen transition, but as the mouse position is in local space, we could use the same one and transform to global. What do you think?

Other than that, i'm thinking that normal pasting should also paste the shapes relative to the new screen position. That is, if i copy a set of shapes that are currently at the center of the screen, move the screen to the left and paste, the pasted shapes should be at the center of the new screen position. In the case the screen has not moved, force the offset to the standard 10,10

Daniferrito:
I went with pasting relative to the screen center position. Now all shapes are pasted relative to the screen's center position, and as it is now done by the paste function itself, it would require extra steps to paste relative to any mouse position.

I notice it is now noticeable the steps in the process of moving shapes to other locations. You can see the shapes disapear, then the location changes, then the shapes appear again. It is probably because my computer is being slow today., but high latency might make it noticeable as well. If its a problem, we might want to disable draw updates until the location transitions completely

client/src/game/units.ts
```diff
@@ -34,7 +34,7 @@ export function l2g(obj: LocalPoint | Ray<LocalPoint>): GlobalPoint | Ray<Global
     const z = gameStore.zoomFactor;
     const panX = gameStore.panX;
     const panY = gameStore.panY;
-    if (obj instanceof LocalPoint) {
+    if (obj instanceof LocalPoint || obj instanceof Point) {
```
Daniferrito: I had to add this, because the functions .add and .substract from LocalPoint (and GlobalPoint as well, but that does not matter) actually return a Point, which was not being picked up here.

Kruptein:
> It would be easier to move shapes relative to the top-left of the screen, as it avoids having to calculate where the center is, but that changes the new position a bit more.

The topleft is in local coords simply (0, 0) :p.  Local coords are literally just your screen starting in the topleft and going to the bottom right. (you can go out of these bounds, but they will not be visible for obvious reasons)

> Other than that, i'm thinking that normal pasting should also paste the shapes relative to the new screen position. That is, if i copy a set of shapes that are currently at the center of the screen, move the screen to the left and paste, the pasted shapes should be at the center of the new screen position. In the case the screen has not moved, force the offset to the standard 10,10

I can agree with this

> I notice it is now noticeable the steps in the process of moving shapes to other locations. You can see the shapes disappear, then the location changes, then the shapes appear again. It is probably because my computer is being slow today., but high latency might make it noticeable as well. If its a problem, we might want to disable draw updates until the location transitions completely

I'll check you code out now to see for myself, but either way it does make sense to no longer do any screen updates from the moment we know we're switching locations.

client/src/game/units.ts
Kruptein: looking at the types for LocalPoint and GlobalPoint this should not be true.  They explicitly override the Point implementation for that specific reason as I want to make sure that at no point the two are wrongly used.  Also subtract does not return a Point at all, but a vector

Daniferrito: Sorry, its only add that was giving me problems.

I know it should be a proper LocalPoint. However, the instanceof was not entering. To see the problem, just checkout the latest commit from this PR and remove this extra condition, and try to copy any shape. It will call the `screenCenter` getter from `store.ts`, which will run `l2g(g2l(this.screenTopLeft).add(halfScreen))`, which will error with the message `Uncaught TypeError: Cannot read property 'origin' of undefined`, which means that `obj instanceof LocalPoint` returned false.

I believe the casting to LocalPoint is only being done at compiler level, but when the actual js runs, a simple Point is returned.

Kruptein:
ok, looking at the code and going over the ideas again I'm a bit conflicted.

When moving shapes to a new location, putting them in the center at that location is exactly what I want so that's perfectly fine.

Pasting shapes in the active location however currently feels a bit inconsistent.  Based on the squaredLength being smaller than 200 (unsure where this number comes from specifically?) you decide whether it's pasted at an offset of 10,10 or at the center of the screen.

I think this is very unclear behaviour for other end users.  I would expect consistent behaviour being either, always at the 10,10 offset or always at the center or another option always at the mouse cursor, but this would involve keeping track of the mouse cursor at all times which is kinda meh.  In the second case you would also always want the center of the screen at the moment of pasting, so there is no need to store the clipboardPosition in between as you can better just calculate the actual screen center at the moment of pasting.

what's your opinion on the above ?

Daniferrito:
The pasting is (almost) always at the screen center. The only time that special offset of 10, 10 is ever used is when the pasted shapes were going to be on top of the original shapes.

Explanation:
`offset.squaredLength() < 200` is equal to `offset.length() < sqrt(200)`, which is equal to `(new Vector(10,10)).length()`, or around `14.1421` I use squaredLength instead of length because it avoids a square root calculation (this is negligible here, but more noticiable when doing many checks or when checking every frame or similar). 

That means that the offset calculated from the center of the screen will be used, except in the case where this offset is less than 14 pixels, in which case a default offset of 14 pixels to the bottom-right will be used.

> In the second case [...], so there is no need to store the clipboardPosition in between

I dont understaund you. To calculate the offset, you need where the original shapes were, and where the new shapes need to be pasted. While you could calculate where the original shapes where from their coordinates (something like a center of gravity of the shapes, or similar), I just do the easy thing and record where the screen was originally, as it will almost always be around the copied shapes. 

This means that if two shapes are copied, one 100 pixels to the right of the center, and the other 100 pixels beneath the center, when pasted, the first shape will be 100 pixels to the right of the new center and the second one will be 100 pixels beneath the new center. That is also true for a single shape, it will not be moved to the center of the screen when pasting.

Another way to look at it, if a shape is at 0, 0, and it is copied, then the screen is moved 20 pixels to the right, and then it is pasted again, the pasted shape will appear at 20, 0, no matter the screen position (only that it moved 20 pixels to the right). The screen could have been at 10000, 10000 (way outside of where the shape was).

Kruptein:
Ok my bad, I misunderstood the code at first glance, should have tried it first. I would add a comment above the 200 though, doesn't have to be very extensive but just something.

client/src/game/units.ts
Kruptein: I can confirm that the LocalPoint info is lost in js after the add call, which is sad :(.  I'm currently looking into ways to fix this.  Maybe as a last resort I literally reimplement the actual code for both Global and Local points, it's not a lot of code anyway and I would really prefer to not having to add the instanceof Point here as I feel it will lead to unforeseen issues in the future.

Kruptein: Ok, got it working! I'm gonna make a quick PR to change the Point implementation and then you should be able to remove the instanceof cast

Daniferrito:
> I would add a comment above the 200 though, doesn't have to be very extensive but just something.

Fair enough, that 200 is basically a magic number, and it deserves an explanation.

Maybe we could move the new Vector(10, 10) to a constant in the class named minimumOffset or something, and check against its squaredLength() instead of 200. The compiler should notice that it is checking against the same thing always and optimize it out as it is right now. However, I dont trust compilers THAT much.

Kruptein:
I'm fine with leaving it as a simple 200, as long as it has the small remark.  Moving it to the constant seems overkill imo.

My Point PR is going through CI so you should be able to merge it soon.
I'm just gonna do a final test of the code and then this should also be fine to merge.

Kruptein:
Hm, the current behaviour when you copy a shape, move the screen slighly and then paste it a couple of times is kinda weird, because of the mutation on the clipboard's x and y.
Doing the offset addition on the `shape` and thus after properly creating the new shape would remedy this, but then the shapes will be pasted on top of eachother if pasted multiple times.

I'm still kinda unsure about the behaviour on panning :D.  I follow the reasoning but it still feels a little bit strange, it might just be the above 'bug' I described.  Another concern is that this could potentially cause a player to get vision in an area (s)he should not be able to see yet due to pasting a token with light there.  Although I'm well aware that can also happen in my original code.  That might be better suited for some other PR though as it's probably more a permission related topic whether or not players should be able to copy/paste freely.

Kruptein:
image
To give some context,
0 is the ogirinal shape, which was pasted 3 times from the same positionm, hence the 10,10 offsets.

I then moved the screen slighlty and pasted again 3 times.  This results in shapes 1, 2 and 3.

Daniferrito:
I see. I will fix this.
What Should have happened:
All shapes around 0 are ok.
All other shapes should be pasted in the same structure as around 0, but around 1.
I will have to cheat a little bit, but it should not be noticeable

Daniferrito:
A one line fix. I expected this to be much uglier.

The same experiment as you did:
Screenshot_2019-09-25_22-32-42

Kruptein:
Nice quick solution.

Did a final test, seems fine to me, it was a long haul, but thanks for sticking through :D

PR #162, opened 2019-09-15
Improved docker image creation and added docker installation steps

Daniferrito:
Added .dockerignore to prevent copying unnecessary files (both improves image creation time and final image size) (the worst offender was that node_modules was being copied to the docker context for building, which thankfully was not being copied to the final image)
Split the Docker image creation in two stages, one for building the frontend and other to distribute the server (previously the frontend had to be compiled separately)
Restructured the Dockerfile to install the dependencies first (as they barely change). This increases the final image size, as it adds a layer, but only by 106 bytes.
Added documentation for how to run PlanarAlly using docker.


Kruptein:
Thanks for all the work today!

I'll take a look at it later as I'm still on holiday currently.

There are a bunch of things I want to change in regards to structure of the repository. One of the things is the current inclusion of the js output files in the repository, it just adds useless merge conflicts and can be trivially created by people checking out the repo anyway.  This means that the exe pipeline has to change to build the js files as well though.

I haven't taken a good look at the docker files in particular, so it's nice to have someone check them.

Daniferrito:
No problem and no rush. This is the last one for today (and probably till next weekend)

For the future, do you prefer a single PR with all features in (but probably separated into commits) or muliple PR with single or very similar features? I did it this way because It is much simpler to accept some of the features and reject/request improvements for the others.

The docker image of all the changes together can be found as daniferrito/planarly in Dockerhub.

Kruptein:
This way is preferred, smaller feature PRs are way easier to reason about and approve, so good job all around.

And yeah development on my end has been going a bit slower recently since my own campaign has been taking some of my free time and I've been working on some improvements to lighting which has caused some blocks on the other todos, so it's nice to have some fresh input :)

docs/installation.rst
```diff
@@ -62,6 +64,52 @@ The binaries are created for each release, which can be found `here <https://git
...
+The server will keep running in the background, reloading after resets
+To stop the server, run `docker-compose down`
+
+Compiling through Docker
+^^^^^^^^^^^^^^^^^^^^^^^^
+Same as the last option, but compiling the image
+directly instead using the precompiled one
+
+Make sure you have a valid Docker installation with docker-compose
+(usually installed sepatedly).
```
Kruptein:
```suggestion
(usually installed separately).
```

docs/installation.rst
```diff
@@ -62,6 +64,52 @@ The binaries are created for each release, which can be found `here <https://git
...
+See :ref:`docker` if you don't know how.
+
+1. Download the latest source code from `here <https://github.com/Kruptein/PlanarAlly/releases/>`_. and open it.
+2. Run `docker build -t planarally .`
+3. Run the newly compiled image with the same instructions as above, but removing the `kruptein/` from the `image` line (the whole line should end up as `image: 'planarally'`)
+
+.. note::
+    Althrough it is not necessary, it is recomended that the folder for the `docker-compose.yml` file is independent from the source folder.
```
Kruptein:
```suggestion
    Although it is not necessary, it is recomended that the folder for the `docker-compose.yml` file is independent from the source folder.
```

docs/installation.rst
```diff
@@ -109,17 +157,62 @@ Download and install the latest python 3 version from `the python site <https://
...
+~~~~~~~~~~~~~~~~~~~~
+
+To install docker, follow the official instructions:
+
+* `Windows <https://docs.docker.com/docker-for-windows/install/>`_ (installation consists in downloading and running the `installer <https://hub.docker.com/editions/community/docker-ce-desktop-windows>`_)
+* `MacOs <https://docs.docker.com/docker-for-mac/install/>`_ (installation consists in downloading and running the `installer <https://hub.docker.com/editions/community/docker-ce-desktop-mac>`_)
+* Linux (Instructions are available for `Ubuntu <https://docs.docker.com/install/linux/docker-ce/ubuntu/>`_, `Debian <https://docs.docker.com/install/linux/docker-ce/debian/>`_, `CentOS <https://docs.docker.com/install/linux/docker-ce/centos/>`_, `Fedora <https://docs.docker.com/install/linux/docker-ce/fedora/>`_)
+
+.. note::
+    If you use linux, you will need to install `docker-compose <https://docs.docker.com/compose/install/>`_ separatedly
```
Kruptein:
```suggestion
    If you use linux, you will need to install `docker-compose <https://docs.docker.com/compose/install/>`_ separately.
```

docs/installation.rst
```diff
@@ -109,17 +157,62 @@ Download and install the latest python 3 version from `the python site <https://
...
+
+To backup the data, make copies to another folder of the following files/folders:
 
 * planar.sqlite: This is the database that contains all the saved sessions and user information.
 * static/assets: This folder contains all the uploaded images for tokens/maps/... during gameplay.
 
 When backed up, you should be able to safely overwrite any of the original PlanarAlly files
 with the newer files from the version you want to upgrade to.
+
+If you are running a manual installation, run `pip install -r requirements.txt` again in case the requiremens have changed.
```
Kruptein:
```suggestion
If you are running a manual installation, run `pip install -r requirements.txt` again in case the requirements have changed.
```

docs/installation.rst
```diff
@@ -109,17 +157,62 @@ Download and install the latest python 3 version from `the python site <https://
...
+
+If the data is lost is not there after upgrading, copy the backups back where you copied them from
```
Kruptein:
```suggestion
If the data is lost or not there after upgrading, copy the earlier backed up files to their original location.
```

docs/installation.rst
```diff
@@ -109,17 +157,62 @@ Download and install the latest python 3 version from `the python site <https://
...
+
+Upgrading with docker
+*********************
+
+Upgrading with Docker is a much safer, but the backups are recommended anyway.
```
Kruptein:
```suggestion
Upgrading with Docker is much safer, but it's still recommended to make backups.
```

Daniferrito:
I agree fully with the changes. Not sure if i just rejected them by clicking the Resolve conversation button. Is it only you who can accept the changes? Or do I merge the changes into my branch somehow?

About changing the route to /planarally, it does not matter much, and we could change it whenever. The only problematic thing would be moving where the permanent files are stored (assets and the database). Moving that would break the easy update from one version to another. However, now would probably be the time to do it.

Actually, it would probably be better to have all the files storing state in a single folder, so only one volume is needed to store all files. Something like a folder named data, with the only contents being the database file and the assets folder. That way we can change this:
        - 'data:/usr/src/app/data'
        - 'assets:/usr/src/app/static/assets'
to this
        - 'data:/planarally/data'
or even better
        - 'data:/data'

However, I dont know the requirements to move the assets folder, and althrough it would work, moving the database inside the assets folder (the relative route comes from a config file if i remember right) seems dirty, as it would mix images with the database file.

Kruptein:
It's the first time I've used the 'suggestions' feature from gh as it popped up as something new to try during me review, I can click the "commit suggestion" buttons, don't know if you can?

Yes we would ideally do this change if we do it rather sooner than later.  I do agree that moving the important files to a single directory is the better approach to make it all a bit more convenient to backup, but I'll need to think about the specifics as moving the save into that folder is indeed dirty, and moving the assets out of the static folder is something that I'll need to update in the serving/routing of those files.  So maybe for now leave it be.

Daniferrito:
That  had to do was go to the files view of this pull request, inspect the changes there and keep pressing add to bundle to all of them. Then click the commit bundle at the top, and that commited your changes to my branch.

So here i changed the folder as requested. For the images I've seen, some are like it was (inside /usr/src), other were like the new way (inside a folder in the root with the name of the app), and some were in the default folders for their non-docker installation. I dont think there is a prefered way.

Adding the container_name to the compose file is not the recomended way, as it limits scalability, but if someone wants to launch multiple instances for some reason, they should edit the dockerfile themselves. On the other hand it removes the weird "change the word docker for whatever your folder is named" and prevents future problems if compose decides to change the default naming scheme.

Dockerfile
```diff
@@ -23,10 +23,10 @@ MAINTAINER Kruptein <info@darragh.dev>
 
 EXPOSE 8000
 
-WORKDIR /usr/src/app
+WORKDIR /planarly
```
Kruptein:
```suggestion
WORKDIR /planarally
```

Dockerfile
```diff
@@ -1,21 +1,40 @@
...
 MAINTAINER Kruptein <info@darragh.dev>
 
 EXPOSE 8000
 
-WORKDIR /usr/src/app
+WORKDIR /planarly
 
-VOLUME /usr/src/app/data
-VOLUME /usr/src/app/static/assets
+VOLUME /planarly/data
```
Kruptein:
```suggestion
VOLUME /planarally/data
```

Dockerfile
```diff
@@ -1,21 +1,40 @@
...
+VOLUME /planarly/static/assets
```
Kruptein:
```suggestion
VOLUME /planarally/static/assets
```

Daniferrito:
I aparently dont know how to write the name of the app. 

Added the changes to the Changelog, but i'm not sure how to proceed with the other PRs. They will probably conflict with dev once the first PR is merged. 

I can think of two solutions:
Write them all now and once every PR is merged, merge back from dev to the branches, solving the conflicts there (they will be easy conflicts, but not automatically resolved)
Write them to each branch as the last step, once they are ready to merged.

Unless you prefer otherwise, I will probably do the first option.

Kruptein:
I prefer the first option anyway :)

I hopefully will have created and merged my eslint/jest/CI changes as well somewhere this week, so that the CI will not just fail on broken lint.

Dockerfile
```diff
@@ -1,21 +1,40 @@
...
+# Build stage for the frontend #
+################################
+FROM node:12 as BUILDER
+
+WORKDIR /usr/src/client
+# Copy first package.json so changes in code dont require to reinstall all npm modules
+COPY client/package.json client/package-lock.json ./
+RUN npm i
+
+COPY . /usr/src
```
Kruptein: It might be better to limit this to the `client` and `server` folders (and in the future exclude the client/tests folder as well).

Daniferrito: It is already added to the .dockerignore. If new folders are added in the future that needs to not be copied, they will need to be added to the  .dockerignore
You can see what actually gets added by running `docker run --rm -it $(docker build -q .) bash` and then inspecting the contents.
From my run, after merging all open PR:
`PA.spec  __init__.py  api     auth.py	 data	 planarserver.py   routes.py  server_config.cfg  t.planar.sqlite  utils.py VERSION  __pycache__  app.py  config.py  models  requirements.txt  save.py    static		 templates`

In any case, the only image that is distributed is the last one. The image that compiles the frontend is a bit more than 1GB in size at the end (node is heavy), but as it only ever lives in developer machines, its not that bad. To reduce that size, the base image could be changed to 12-alpine, which [has problems with some libraries](https://hub.docker.com/_/node/), but seems to work for this project (docker recognizes the built static files as the same so the final image has the same checksum). It reduces the build image size to 340MB. 

We could also not copy the full server folder to the build images, as it is not strictly needed, but copying it all is much simpler.

Kruptein: Right, yeah my bad, thanks for clearing that up.

Kruptein:
I should add a docker build step to CI to make sure docker images can still build as well, but this seems good to go, thanks for the work.

PR #165, opened 2019-09-17
Feature/eslint+jest+ci

Kruptein:
This PR introduces eslint and jest as well as an update to the CI pipeline.


Daniferrito:
This breaks a bunch of things for PS #161, but I mainly agree with the changes.

The getRef didnt look right to me, althrough I dont totally agree with the EventBus remplacing it. Probably I'm just not used enough to this kind of one-sided events.

I would sugest, however, to move the inInitiative() method to InitiativeStore. Also, addInitiative() already checks if the shape is already in the initiative, no need to check before adding. No big deal since that part of the code is rewritten in #165 (we still need a separate inInitiative() for getInitiativeWord())

Kruptein:
Hey, thanks for taking a look.

Yeah the getRef and its current solution are both not ideal to me either.  I seem to be missing something in regards to Vue communication in this case, if you have any other suggestions feel free to give more input.

You're right that its redundant to check inInitiative on addInitiative, I'll move the file to the store as well  (but not to the store class itself)

Kruptein:
I'm gonna move forward and merge this PR in its current state so that the CI works properly and I can continue work on my lighting improvements amongst others.

If you do have some input/suggestions to make the getRef/EventBus stuff cleaner, feel free to open an issue or comment on this PR.

Daniferrito:
I was merging this into my PRs, and i noticed a bug. After investigating, i found it was not my branch but this update.

In particular, it broke the "Create new location" button:
`TypeError: Cannot read property 'prompt' of undefined`

from line 176 in `Game.vue`:
``        (<Game>this.$parent.$parent).$refs.prompt.prompt(`New location name:`, `Create new location`).then(
``

Kruptein:
I can confirm this, the line in menu.vue:176 should be

`(<Game>this.$parent).$refs.prompt.prompt...` instead

Kruptein:
Waiting on CI approval for the bugfix PR to merge

Kruptein:
Merged

Issue #130, opened 2019-02-15
Keybinding screen


Kruptein:
Some screen that at the very least shows all available key shortcuts and optionally alter some of these to custom choices.

Issue #67, opened 2018-09-04
Find a way to recover from "scrolling away" from the map


schemen:
Hi!

When scrolling out of a map and in again, let's say accidentally  - you have no chance to find the map again as it is basically an infinite canvas.
There should be a way to go back to point 0 or a "deployed assets" view where you can go to said asset.

As a quick report also, I cannot drag&drop maps into the canvas on the latest vue commit

Cheers!

Kruptein:
Hi, this is indeed a very valid problem that I've had happen to me multiple times as well, it's good to have this on the issue tracker as I often forget to work on it ;)

And indeed the drag and drop support was temporarily disabled while I was working on other things but is one of the small amount of items still left to do.

schemen:
Do you have any idea on how to approach this problem? 
Maybe limit the canvas or use a similar approach of "Move player here"  but "move to asset xy"

Kruptein:
I was thinking about introducing a system to mark locations either by right right click or some dedicated tool.   And then giving the ability to jump to these points, and then providing a basic (0, 0) point or something.

Does this sound good for you ?

schemen:
Mh, sounds good, provided a dropped asset would automatically represent a marked locations where you can jump to I guess!

Kruptein:
That would quickly clutter things though.  I would just go with providing the location that you first see when creating a new map as a default 'marked location'.  And giving the option to add some manually.

schemen:
Ah, like a center beacon? And when you create them map there it should be no problem to move back to it?

Kruptein:
Yes indeed.

So you can manually add beacons and go to them at any time.

And on top of that a 'default beacon' should be provided, this could be the center/origin point of the canvas, or could be the very first shape added.

schemen:
Maybe the first object on the map layer serves as initial beacon?

LDeeJay1969:
I found myself in this situation quite a bit (as a noob), so I edited keyboard.ts and added
```
        } else if (event.key === "r" && event.ctrlKey) {
            // Ctrl-r - Re-center/reset the viewport
            gameStore.setPanX(0);
            gameStore.setPanY(0);
            sendClientOptions(gameStore.locationOptions);
            window.location.reload();
```
[...]

PR #178, opened 2019-09-25
Added Keybinds window

Daniferrito:
Implements #130

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,12 @@ All notable changes to this project will be documented in this file.
 - Added the default right click menu to all tools that didnt have it
 - Added colour in the location bar to show current location
 - Added keybinds window
+- Added Excape keybind that closes all modals and deselects all shapes
 
 ### Changed
 
 - Pasted shapes are now pasted relative to the screen position
+- Windows now reorder so that the most recent interacted window is on top
 - [tech] Mousemove events are now throttled, so that they don't fire a gazillion events.
 - [tech] tslint swapped out for eslint
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 
 - Added option to set custom units of length
 - Added ping tool
+- Added keybinds window
 
 ### Changed
 
--- a/client/src/core/components/modals/closeableModal.vue
+++ b/client/src/core/components/modals/closeableModal.vue
@@ -64,7 +64,7 @@ export default class CloseableModal extends Vue {
     beforeDestroy() {
         EventBus.$off(`${this.title}.Toggle`);
         EventBus.$off(`${this.title}.Open`);
-        EventBus.$off(`${this.title}.Close`)
+        EventBus.$off(`${this.title}.Close`);
     }
 
     handleClick(event: { target: HTMLElement }) {
--- a/client/src/core/components/modals/closeableModal.vue
+++ b/client/src/core/components/modals/closeableModal.vue
@@ -1,0 +1,103 @@
+<template>
+    <Modal ref="modal" :visible="visible" :colour="'rgba(255, 255, 255, 0.8)'" @close="close" :mask="false">
+        <div
+            class="modal-header"
+            slot="header"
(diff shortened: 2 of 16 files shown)


Daniferrito:
This is just a work in progress. Not sure if I can set that flag or not.

Here is how it looks so far (I changed the title of the window to "Keybinds" since the screenshot).

Screenshot_2019-09-25_21-26-11

Issues remaining:
* Add a more unified way of adding this kind of window menus (maybe not needed if there are no more planned).
* Sort the windows correctly. Right now, "Keybinds" is always on top of "DM Options". Not a big issue, but the most recent window could be moved to the top (by changing its z-index, i believe).
* Add whatever keybinds are there that i did not add. I got this ones from the keyboard.ts file, but i'm not sure it is all of them
* Allow the possiblity to edit the keybinds. Maybe not all of them (I would still keep the copy, paste and cut non-editable)
* Fix the notes at the top of the movement section.

Kruptein:
Already looks quite good!

In the future we might want to move some things to different tabs in that window, but for now it's more than fine.

I would indeed like to allow changing keybindings at one point, but this also needs server side capabilities to save the changes. and that can be added at a later instance.

Just have to check the code and play with it myself, but already seems merge-able

Kruptein:
Missed the work in progress note,  I can during PR creation click on a dropdown next to the 'create pull request' button to select 'create draft pr', unsure if that is also visible for you.  Either way, just comment when it's finished

Daniferrito:
Oh, I created it from outside github, so i did not have that option. I will try it next time i create a PR (which should be soon, I have my eyes on an extra context-menu option to go to the nearest shape, to find the relevant part of the map again)

Kruptein:
Related to that there is [this](https://github.com/Kruptein/PlanarAlly/issues/67#issuecomment-418415780) which I still intend to add at some point.  Basically a way to give quick-access to certain points.  I'm unsure if that's something we want to add to the context menu, or rather have it in the sidebar menu like notes currently are.

Daniferrito:
Ok, more work done. I still dont like how the notes at the top of the movement section look, but I'm not sure how to change it, and The two other people that have seen it didn't see any problems.

A small bug that this introduces is that the Labels window is shown underneath the shape properties window that spawned it, as somehow the click event from Modal is triggered last no matter how I try to delay the moving to the top of the Labels window.

By the way, (as it is not in the main body of the PR), I just added a ModalsStore, that is in charge of reordering all the modals, by changing their z-index (was 9998 for all of them, it is now 8999 to 8000 in order of most-recent-interacted-with to least). I will probably change it so it also handles which modals are visible (it already knows, but currently visibility is handled by each individual modal).

Finally, I also extracted all the common logic of Closeable Modals into CloseableModal. It means that there are more EventBus subscriptions, as all Modals subscribe to its own Toggle, Open and Close, even if noone will ever trigger send them. I dont think this is a problem, though.

I will probably also add a key-bind for Escape that closes all modals, closes the top-left menu and clears the selection.

Kruptein:
Haven't checked the code yet, but quickly ran it already.  Some input:

* I like the modal stacking change, seems to work pretty intuitive so far.
* Escape closing all modals and menu's feels a bit weird.  I would expect it to close the top-most modal/dialog and if no modal present perhaps the sidebar as well, although I would move that to another key (and have a dedicated key for the locations panel as well).
* I don't like the current way the keybinds option is presented in the sidebar, I understand this is mostly due to my current styling rules, but this ideally looks cleaner in the final version
* I would prefer to simply use 'keybindings' instead of keybinds in the sidebar as well as for the title of the keybinding dialog
* The properties dialog in general is broken style-wise, it used to not be transparent at all, which it does now, hence some input fields looking strange.

Just as an aside from feedback from another DM (this does not have to be changed in this PR, but I would like to address it),  I would drop the shift and only use ctrl for the multi selection.  As shift is also commonly used to move assets through movement-blocking features, this can cause for accidental group movements when one just wants to move another shape and was already pressing shift.

Kruptein:
Hey, just checking in to see if you still have time to look at this (if you are still using PA).
Otherwise I'll gladly continue work on this PR

PR #182, opened 2019-10-05
Feature/multiline

Kruptein:
This PR removes the multiline and instead merges its functionality in the polygon shape as it had many overlaps.

Additionally polygons can now be open as well.


Daniferrito:
I agree with the changes, but I have two issues currently:

The interface is confusing. I see references in the code to text asking if this is a filled or not polygon, but i cannot find them.

I'm missing the possibility to draw a multiline without the last edge. This might be the openPolygon part, but i cannot find the way to toggle it.

Kruptein:
The polygon draw tool has a "Closed polygon?" toggle.

The UI Is pretty horrible currently that's still WIP

Kruptein:
The traditional multiline was used only for the paintbrush, which has a line width property, hence the need for the polygon to also have it.

I'll probably also add a brush width / line width  slider or input field for the polygon tool itself.

Kruptein:
The last commit adds the brush size UI also to the polygon tool.

There are 3 ways of creating a "multiline" now.

1) paint brush tool
    This allows for free form drawing with the mouse but is under the hood implemented as a multiline
2) polygon tool with closed option
    This is the original polygon, left click to add points, right click automatically closes the polygon
3) polygon tool with closed option unchecked
    This behaves the same, left click to add points, right click will "complete" the shape, but not add a connection between the last point and the first point.

PR #192, opened 2019-11-14
Changed icons to words

Goury:
Because some of my autistic players can't read icons, but can read words just fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file.
 - All shapes on the FOW layer are now invisible while not on the FOW layer.
 - Circle borders (including basic tokens) are now inset, so that they fit within their squares.
 - Basic token text scaling has been changed slightly to have more uniformly sized characters.
+- Some UI icons are now words.
 
 - [tech] Mousemove events are now throttled, so that they don't fire a gazillion events.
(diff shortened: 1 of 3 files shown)


Kruptein:
Ok so when I read your mail about text instead of icons I thought the in-game icons where not clear (e.g. vision related icons), I wasn't expecting these icons to be the point of contention.

I personally really like the simplified/minimal button approach, so I'm not keen to change it as the default.

That said I see the validity of your users' concern and I'm willing to make it a customisable user setting using localstorage that defaults to the icons but gives the option in the settings to change to text.

So what I would like to do for this is first pick up the user settings again, which is an ongoing branch on my end that I need to finish and if that's merged further look at this PR.

You can always stash this change so that you can reapply it for your own server in the meanwhile when changes to my code happen.

Does this work for you ?

Kruptein:
Also I would center the create button in the session overview.

Goury:
I don't mind having this change just in my fork, but defaults should totally be administrator-configurable.    
I believe players who already failed an ability check to understand meaning of icons won't succeed on finding a way to adjust settings.

Some of my players seem had rolled –20 for wisdom, so I'd prefer to keep all perception checks DC as low as possible and to have as little of those as possible.  
As for my DMs, they are good.

Kruptein:
The UI of these places has long since changed.

PR #207, opened 2020-02-08
Recover from "scrolling away" from the map

LDeeJay1969:
Resets the viewport of the client back to 0,0

### DIFF
--- a/client/src/game/events/keyboard.ts
+++ b/client/src/game/events/keyboard.ts
@@ -68,6 +68,12 @@ export function onKeyDown(event: KeyboardEvent): void {
             event.preventDefault();
             event.stopPropagation();
             gameStore.toggleUI();
+        } else if (event.key === "r" && event.ctrlKey) {
+            // Ctrl-r - Re-center/reset the viewport
+            gameStore.setPanX(0);
+            gameStore.setPanY(0);
+            sendClientOptions(gameStore.locationOptions);
+            window.location.reload();
         } else if (event.key === "c" && event.ctrlKey) {
             // Ctrl-c - Copy
             copyShapes();
--- a/client/src/game/events/keyboard.ts
+++ b/client/src/game/events/keyboard.ts
@@ -68,12 +68,12 @@ export function onKeyDown(event: KeyboardEvent): void {
             event.preventDefault();
             event.stopPropagation();
             gameStore.toggleUI();
-        } else if (event.key === "r" && event.ctrlKey) {
-            // Ctrl-r - Re-center/reset the viewport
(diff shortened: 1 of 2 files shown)


client/src/game/events/keyboard.ts
```diff
@@ -68,6 +68,12 @@ export function onKeyDown(event: KeyboardEvent): void {
             event.preventDefault();
             event.stopPropagation();
             gameStore.toggleUI();
+        } else if (event.key === "r" && event.ctrlKey) {
```
Kruptein: Ctrl+R is a common keybinding to reload a webpage and I would thus prefer not to override it, as it could be that a user wants to actually hard reload the page if something seems wrong.  I would either choose a different combo (be it a different modifier or letter) or even just use the R key without any modifier.

client/src/game/events/keyboard.ts
```diff
@@ -68,6 +68,12 @@ export function onKeyDown(event: KeyboardEvent): void {
...
+            // Ctrl-r - Re-center/reset the viewport
+            gameStore.setPanX(0);
+            gameStore.setPanY(0);
+            sendClientOptions(gameStore.locationOptions);
+            window.location.reload();
```
Kruptein: This unnecessarily reloads the entire page including all the data transfers and bootup logic.
What you want to do instead is tell the canvas to rerender, which you do by telling the various layers that they are no longer valid with: `layerManager.invalidate()`

client/src/game/events/keyboard.ts
LDeeJay1969: You're right. But it's difficult to find one that's really unique and maybe  not in use by other apps. In the most ideal situation you'd let the user pick a combo. I wouldn't use a non-modified letter since that might interfere with typing actions during 'normal' operations.

Kruptein: I understand the sentiment, but this is something that should already be properly handled.  The 'd' key deselects the current selection, but not if the user is typing in a textfield/textarea as can be seen in lines 21-23

client/src/game/events/keyboard.ts
LDeeJay1969: Ok, I've updated the code. CTRL-0 (zero) is now the hot-key and the layermanager invalidated. No more reloads. BTW, if I want to write some settings to the db, how would I approach his? (Think user settings/key settings etc) This might also help with #178. I can have a look at this combination if you want.

Kruptein:
This looks good to go, one last thing I forgot to mention is to update the CHANGELOG file in the root of the repository with your change.  Feel free to add something like [contributed by LDeeJay1969 or something at the end of the line.

Issue #144, opened 2019-05-07
Support Touch devices/Touch interface


schemen:
I would love to be able to use this app on my iPad to cleanly use everything on a single device. Any chance to have a look at that?

Cheers!

Kruptein:
Hey, I understand the usecase and would like to get this working but I don't have any tablet device myself.  I do believe I can emulate it with chrome devtools, I'll see how that goes.

schemen:
No worries! I'll also be testing it all the time when you need help - I can deploy a testing branch onto dndbox if you want :)

ZachMyers3:
I'm using PA on a touch device, so I'd like to look at implementing it.

It looks like there's a [vue2-touch-events](https://www.npmjs.com/package/vue2-touch-events) library that would make this possible. I'll play around with it on my fork and see what I can do.

Kruptein:
That would be great!

You could also look into the general touch events.(https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent).  These should probably also be attached to the current mouse event functions

ZachMyers3:
I took your advise and dug into the TouchEvents. 

It looks like every component that has a ```MouseEvent``` event handler should have a ```TouchEvent``` handler as well. I'm still getting familiar with the codebase but I think this is the work list I was able to come up with.

* [tool.vue](https://github.com/Kruptein/PlanarAlly/blob/3edc8180a3c55d629c7e89397486ce5132df27a9/client/src/game/ui/tools/tool.vue)
* [tool.vue](https://github.com/Kruptein/PlanarAlly/blob/1a36442914329510a73aa08bd0d8385737331def/client/src/game/ui/tools/tools.vue)
* [pan.ts](https://github.com/Kruptein/PlanarAlly/blob/1a36442914329510a73aa08bd0d8385737331def/client/src/game/ui/tools/pan.ts)
* [map.vue](https://github.com/Kruptein/PlanarAlly/blob/1a36442914329510a73aa08bd0d8385737331def/client/src/game/ui/tools/map.vue)
* [ping.ts](https://github.com/Kruptein/PlanarAlly/blob/1a36442914329510a73aa08bd0d8385737331def/client/src/game/ui/tools/ping.ts)
* [ruler.ts](https://github.com/Kruptein/PlanarAlly/blob/1a36442914329510a73aa08bd0d8385737331def/client/src/game/ui/tools/ruler.ts)
[...]

PR #213, opened 2020-02-20
Add touch events for #144

ZachMyers3:
This adds support for touch events within the tools in ```game.vue```. Refactoring of some existing events were done for code re-use within the touch events. 

A swipe feature was built within ```game.vue``` using vue2-hammer to allow for the ability to pinch zoom the map.


ZachMyers3:
Having some trouble getting this packing to be recognized by the CI. I'm able to use the package fine locally, any thoughts? I see there's a shims folder for declaring modules but don't see anything that pulls in declared types like vue2-hammer does.

Kruptein:
It also fails in my local build.  Run `npm install --save @types/vue2-hammer` to add typings and you should be good to go :)

client/src/game/ui/tools/pan.ts
```diff
@@ -6,29 +6,52 @@ import { sendClientOptions } from "@/game/api/utils";
...
-import { getMouse } from "@/game/utils";
+import { getLocalPointFromEvent } from "@/game/utils";
 
 @Component
 export default class PanTool extends Tool {
     name = "Pan";
     panStart = new LocalPoint(0, 0);
     active = false;
 
+    panScreen(l: LocalPoint): void {
```
Kruptein: Maybe rename `l` to `target` for a bit more clarity.

client/src/game/ui/tools/pan.ts
```diff
@@ -6,29 +6,52 @@ import { sendClientOptions } from "@/game/api/utils";
...
+    }
+
+    onTouchEnd(_event: TouchEvent): void {
+        this.active = false;
+        sendClientOptions(gameStore.locationOptions);
+    }
+
+    onTouchMove(event: TouchEvent): void {
+        if (!this.active) return;
+        // get the point of the mouse event
```
Kruptein: Slightly redundant comment :)

client/src/game/ui/tools/select.vue
```diff
@@ -45,7 +45,8 @@ export default class SelectTool extends Tool {
         this.selectionHelper.globalCompositeOperation = "source-over";
         gameStore.setSelectionHelperId(this.selectionHelper.uuid);
     }
-    onMouseDown(event: MouseEvent): void {
+
+    onDown(lp: LocalPoint, gp: GlobalPoint, event: MouseEvent | TouchEvent): void {
```
Kruptein: Drop the gp argument here and actually do the l2g conversion you're doing now in both onMouseDown and onTouchDown in this function

ZachMyers3:
I made all the changes you've suggested, and CI is working as intended. On my initial commit I did run that install command you suggested, but maybe I didn't run with --save to add to package.json.

Let me know if you have anything else!

client/package.json
```diff
@@ -9,6 +9,7 @@
     "clean": "rimraf ../server/static/js/ && rimraf ../server/static/css/"
   },
   "dependencies": {
+    "@types/vue2-hammer": "^2.1.0",
```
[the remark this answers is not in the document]
ZachMyers3: Changed in latest commit

ZachMyers3:
VueHammer is able to tweak the threshold of the pinch, but I was having trouble getting any threshold other than the default to function any better. 

```typescript
// main.ts
VueHammer.config.pinch = {
    threshold: 1,
}
```

As far as the direction, that's just me getting the directions mixed up and I've changed that in my latest commit. 

I'm fine with making Pan the only pinch-enabled zoom feature, it works cleanest with Pan anyway. On top of that it works fair well "as-is" (i.e. there wouldn't necessarily be a problem with firing the touchdown and pinch events) if we went that route.

I'll do a little more digging into hammer.js, maybe there's more ability to tweak the gesture than I currently am aware.

client/package.json
```diff
@@ -49,7 +48,8 @@
     "rimraf": "^3.0.0",
     "ts-jest": "^24.1.0",
     "typescript": "~3.7.5",
-    "vue-template-compiler": "^2.6.10"
+    "vue-template-compiler": "^2.6.10",
+    "vue2-hammer": "^2.1.2"
```
Kruptein: Slight whoopsie :D  It's the `@types/vue2-hammer` dependency that has to go in the devDependencies as it provides nothing at runtime.  The `vue2-hammer` dependency has to be in the normal dependencies as it actually provides functionality.

Kruptein:
Yeah, I would for now make it so that only pan has pinch-zooming, but I would still like to see it be either pan or either pinch and not both at the same time.

ZachMyers3:
I'll have the event fire through ```game.vue``` into the pan tool and handle it in there.

That would also make it relatively easy to not fire the touchdown panning either.

client/package.json
ZachMyers3: It looks like CI worked without it being in the standard dependencies, do you want me to revert it?

Kruptein: Yes :D and move the @types to the devDepdencies

ZachMyers3:
I ended up branching the pinch changes and sidelined them for this PR.

Like you said, it was a little too clunky and the zoom slider works just fine with touch controls as it stands now.

I think if I really wanted to get this clean I'd have to go with my original approach and determine a pinch inside of the existing ```TouchEvent``` frameworks. Hammer.js adds a little more overhead than I'd like to see in this use-case.

ZachMyers3:
I think it's (finally (hopefully)) ready to go!

Kruptein:
The only last remark I have is that the package-lock is marked as completely different due to your version having 2 space indent whereas I use 4 space indent.

Just reset the file I suppose and then it's good to go

ZachMyers3:
When I reset the file locally it uses two space indents instead of four, I just copied from the current dev branch since I didn't adjust any dependencies.

PR #214, opened 2020-02-20
Markers #67

LDeeJay1969:
For issue #67 markers are added: any shape can be marked as marker by right-clicking and selecting 'set as marker' from the context. From the menu on the left this name can then be selected and will become the centre of the screen. From this menu the marker can also be removed.


LDeeJay1969:
Apologies for the messy commit. I forgot to update my locate repository so committed to an old dev at first.

server/api/socket/location.py
```diff
@@ -51,7 +52,24 @@
...
     )
-
+    sorted_initiatives = [
+        init.as_dict()
+        for init in Initiative.select()
+        .join(Shape, JOIN.LEFT_OUTER, on=(Initiative.uuid == Shape.uuid))
+        .join(Layer)
+        .where((Layer.location == location))
+        .order_by(Initiative.index)
+    ]
```
Kruptein: This should not be here anymore

server/models/campaign.py
```diff
@@ -166,6 +167,21 @@ def as_dict(self):
             self, recurse=False, exclude=[Note.room, Note.location, Note.user]
         )
 
+class Marker(BaseModel):
```
Kruptein: I see some issues here.  You declare both room and location.  This can only lead to potential bugs (e.g. a marker with room=a and location=l  where l is a location in room b).  You have access to room via the location refkey, so there is no need for a direct link to room.

Coincidentally there is actually also no real reason to have a direct link to location.  What I think you actually want, is to have the uuid be a reference to a shape directly, as a Marker in its current form is always associated with a particular shape.  Take for example ShapeLabel in shape.py, it has 2 direct foreign keys.  One to shape and one to label.  What we want here ultimately is 1 reference to shape and 1 to user, nothing more.

server/api/socket/location.py
```diff
@@ -51,7 +52,24 @@
...
+        .join(Shape, JOIN.LEFT_OUTER, on=(Initiative.uuid == Shape.uuid))
+        .join(Layer)
+        .where((Layer.location == location))
+        .order_by(Initiative.index)
+    ]    
+    await sio.emit(
+        "Markers.Set",
+        [
+            marker.as_dict()
+            for marker in Marker.select().where(Marker.user == user)
```
Kruptein: This will return all the markers you own irregardless of the session/location you're in.

client/src/game/api/events.ts
```diff
@@ -96,6 +96,9 @@ socket.on("Position.Set", (data: { x: number; y: number; zoom: number }) => {
 socket.on("Notes.Set", (notes: Note[]) => {
     for (const note of notes) gameStore.newNote({ note, sync: false });
 });
+socket.on("Markers.Set", (markers: Marker[]) => {
```
Kruptein: What you receive here is not in fact a Marker, but a Pick<Marker, 'uuid'> or rather just a string[].  The Marker.name aspect does not come from the server.

That said, I have some remarks on the Marker.name below

client/src/game/comm/types/general.ts
```diff
@@ -67,3 +67,9 @@ export interface Note {
     text: string;
     uuid: string;
 }
+
+export interface Marker {
+    uuid: string;
+    name: string;
```
Kruptein: I don't fully see the need for a separate name field here.

You have the uuid so you can always retrieve the shape name at any time.  If anything, storing the name in 2 different locations is bound for some sync issues.  I can already tell that at this point, renaming a shape will in fact not rename the marker until you refresh the page.

I think you want to keep the markers as a simple string[] array containing uuids in the current marker idea.

client/src/game/ui/menu/menu.vue
```diff
@@ -68,6 +68,25 @@
...
+                                        <tr v-for="marker in markers" :key="marker.uuid" style="cursor:pointer">
+                                            <td style="white-space: nowrap;" width="99%" @click="jumpToMarker(marker)">
+                                                {{ marker.name || "[?]" }}
+                                            </td>
+                                            <td style="white-space: nowrap;" @click="delMarker(marker)">
+                                                <i class="far fa-minus-square"></i>
+                                            </td>
+                                        </tr>
+                                    </tbody>
+                                </table>
```
Kruptein: I'm not super fond of <tables> :p If you can, rework it to a flexbox or grid, if you're not that familiar with it, I don't mind doing this as it shouldn't take a lot of effort

server/api/socket/location.py
LDeeJay1969: Ooops.... paste errror...

client/src/game/ui/menu/menu.vue
LDeeJay1969: Maybe not but this was twofold: firstly, when I used your approach, when making a lot of markers and notes, the menu bar gets a scroll bar. Scrolling would make the text and the icons move independently from each other (it still happens with the + for the notes). Using tables created a more 'sturdy' design which moved them together. It might have to do with the width that changes when adding the scrollbar but didn't want to mess with it too much.
Secondly it feel this gives the programmer a slightly more clear view about the design rather than having a lot of divs and having to figure out which class does what. But that's entirely personal.

server/api/socket/location.py
LDeeJay1969: And that's what you want, right?

client/src/game/comm/types/general.ts
LDeeJay1969: This was kinda by design.... I thought that if you want to mark shapes with similar names you can differentiate them by naming them, marking them and then renaming them. You marker will then have a different, more distinctive name. But you're right, it doesn't have to be stored and might lead to confusion.

LDeeJay1969: On that note, if you want to make a more elaborate system where you want to 'share' makers etc, you might want to keep a name. And access-info (who has it), But realizing that, you might want to change the shape-focused marker to a name-to-shape focused marker so you can share the same shape with different marker-names....  tricky one. 

I'll remove it for now, it can always be added afterwards...

LDeeJay1969: And in doing so I just realized I wanted to keep them in an object to keep track of the two. In the menu on the left I display the name but keep track of the uuid. This way you can add markers with similar names but not with a duplicate uuid.

server/api/socket/location.py
LDeeJay1969: No you don't ..... I'll add them...

client/src/game/comm/types/general.ts
LDeeJay1969: This will be reset when you refresh though. If you want to keep them, you'll need to store them in the db as well.

Kruptein: As you point out yourself, there is actually no mechanism to rename a marker and even if there was it would be overwritten on load by the shape name, unless you remove that aspect and set the default name for a new marker to the shape it is attached to.

You don't need to track the name in an object in order to display the name.  Just request the shape from the layermanager and show that name instead.


tl;dr  Either you offer true marker renaming where it is not associated with the shape itself or you only track the shape uuid as keeping track of the name on top of this is just bound to fail.

client/src/game/ui/menu/menu.vue
Kruptein: Using css grid is just simpler imo.  The 99% width and nowrap whitespace stuff feels wrong.

There is no absolute right/wrong here as this is just my preference ;)

client/src/game/comm/types/general.ts
LDeeJay1969: That's what I don't get. Once it's pushed to array to show the name, I have no way of knowing from which shape-uuid it came unless I keep them together. Besides, once you click on the name, you want to know to which shape-uuid to go to. So you need to keep a relation between the two.

Kruptein: `layerManager.UUIDMap.get(marker).name` will give you the name of the shape with uuid=marker

server/models/campaign.py
LDeeJay1969: I'm so annoyed (with myself) at the moment. I had the process working with strings instead of objects, then when hitting refresh, all markers were read as objects. So I changed the class to return a string instead of objects, now that works, now I can't write to the db. Whenever I want to, the server responds with "Task exception was never retrieved" on executing the `Marker.get_or_none(uuid=data)` where I'm sure data is a string (according to the log), not an object but when I change data to data.uuid, I get "'str' object has no attribute 'uuid'". That makes sense but I have no clue how to change the marker.py contents to read the db as string instead of object or to get rid of the warning.

LDeeJay1969: Nevermind...... just a contraint on the table that I completely overlooked. All's well again...

LDeeJay1969:
Changed from object to string, resolved the delete marker on delete shape and reworked the menu from table to flex. Still to do is set locations as marker. This can be added later though...

server/api/socket/location.py
Kruptein: There is still extra filtering required.  Imagine you have two completely different sessions in which you use markers.  This will return all markers from both sessions combined, which you wouldn't want.

server/models/campaign.py
```diff
@@ -166,6 +167,15 @@ def as_dict(self):
...
+    uuid = TextField(primary_key=True)
```
Kruptein: I would make the uuid a ForeignKeyField of Shape, as that's what it ultimately is.

Also, right now you'll get a rejection from the database if another user tries to set a marker on a shape someone else has already marked as uuid is a primary key.

client/src/game/shapes/utils.ts
```diff
@@ -156,6 +156,7 @@ export function deleteShapes(): void {
         l.removeShape(sel, SyncMode.FULL_SYNC);
         EventBus.$emit("SelectionInfo.Shape.Set", null);
         EventBus.$emit("Initiative.Remove", sel.uuid);
+        gameStore.removeMarker({ marker: sel.uuid, sync: true }); //delete marker.... even if it doesn't exist.
```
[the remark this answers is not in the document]
LDeeJay1969: Moved... 
Should there be a check to see whether the shape is marked or can I just try to remove the marker, even it might not exist? If I need to check first, I need to access (or create) the marker array which layer doesn't have direct access to, which means more references and arrays.

Kruptein:
Let me know when I can review this again.
Also some small advice, small commits are great, but try to push them in larger batches ;) I get a lot of emails currently :D

LDeeJay1969:
Apologies for the commits, whilst trying to rearrange the code, I run into some issues....

I've checked for sessions now so you don't delete markers between different sessions. All works fine but I have to refresh the browser to see the menu. When I just switch sessions, the menu (cog-icon) won't work until I refresh the browser. When I remove `return layerManager.UUIDMap.get(marker)!.name;` from menu.vue it works ok, apart from showing [?] instead of the name (understandably). I can't figure out where this behaviour is originated.

Kruptein:
I'll try to give this a look in the weekend, so that we can hopefully get this working when I want to release 0.12.

I want to focus my efforts this week in one last ticket so that I can make a 0.11 release this week.

I haven't tested anything, but your `When I remove return layerManager.UUIDMap.get(marker)!.name;` sentence seems to imply to me that the returned value is not in fuck defined (the !),  Are there any console logs that can help you further until I take a more in depth look?

LDeeJay1969:
I can switch between sessions now (the !  was indeed the issue). But the solution for this, poses the next problem: when loading the page, all is fine (all markers are named OK). When switching between sessions, _all_ markers are listed (also the ones from other sessions) and the ones not belonging to the current session are shown as [?].  So loading the page has the right query, switching sessions doesn't...

LDeeJay1969:
Finally .... I figured out I didn't reset the list properly between switching sessions so once that was solved it works a treat! I have thoughts for adding markers on locations as well but I'll make a new issue for that one.

Kruptein:
Glad to hear you got it sorted out, I release 0.19 earlier , so I should have some more time now to properly review this PR again

LDeeJay1969:
The only thing I haven't working yet is the marker as foreign key of Shape. Once I tried to change that, a whole lot of errors emerged on the server side and it wouldn't start anymore until recompiled so I have to dig deeper into that. Not sure if that's a huge issue at the moment though.

LDeeJay1969:
Finally got the foreign key working. Probably not the most elegant(?) but it works. Since I had to switch to a as_dict (what Shape is) I struggled getting just a string back. But got a solution in the end. Let me know what you think.

server/models/campaign.py
```diff
@@ -166,6 +167,20 @@ def as_dict(self):
             self, recurse=False, exclude=[Note.room, Note.location, Note.user]
         )
 
+class Marker(BaseModel):
+    from .shape import Shape
+
+    uuid = ForeignKeyField(Shape, backref="markers", on_delete="CASCADE")
+    user = ForeignKeyField(User, backref="markers", on_delete="CASCADE")
+    room = ForeignKeyField(Room, backref="markers", on_delete="CASCADE")
```
Kruptein: Any particular reason why you chose to use Room and not Location?

server/save.py
```diff
@@ -294,7 +294,9 @@ def upgrade(version):
             db.execute_sql(
                 'INSERT INTO layer (id, floor_id, name, type_, player_visible, player_editable, selectable, "index") SELECT _layer.id, floor.id, _layer.name, type_, player_visible, player_editable, selectable, _layer."index" FROM _layer INNER JOIN floor ON floor.location_id = _layer.location_id'
             )
-
+        db.execute_sql(
+            'CREATE TABLE IF NOT EXIST "marker" ("id" INTEGER NOT NULL PRIMARY KEY, `uuid` TEXT NOT NULL , `user_id` INTEGER NOT NULL, `room_id` INTEGER NOT NULL, FOREIGN KEY(`uuid`) REFERENCES `shape`(`uuid`) ON DELETE CASCADE, FOREIGN KEY(`room_id`) REFERENCES `room`(`id`) ON DELETE CASCADE, FOREIGN KEY(`user_id`) REFERENCES `user`(`id`) ON DELETE CASCADE)'
+        )
```
Kruptein: This needs to move to a separate elif clause

LDeeJay1969:
I'm so sorry for this mess.....  it's  very steep learning curve for me. 
But I think the penny finally dropped where I went wrong. Will update again.

client/src/game/store.ts
```diff
@@ -246,6 +248,36 @@ class GameStore extends VuexModule implements GameState {
         if (data.sync) socket.emit("Note.New", data.note);
     }
 
+    @Mutation
+    newMarker(data: { marker: string; sync: boolean }): void {
+        let mrk = "";
+        if (typeof data.marker === "object") {
```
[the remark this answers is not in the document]
LDeeJay1969: Resolved this now.... turned back into a proper string. Went wrong on the server side.

client/src/game/store.ts
```diff
@@ -246,6 +248,36 @@ class GameStore extends VuexModule implements GameState {
...
+            mrk = String(Object.values(data.marker)[0]);
+        } else {
+            mrk = data.marker;
+        }
+        const exists = this.markers.some(m => m === mrk);
+        if (!exists) this.markers.push(mrk);
+        if (data.sync) socket.emit("Marker.New", mrk);
```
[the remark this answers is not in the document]
LDeeJay1969: Again, went wrong at the server side. Resolved now.

client/src/game/store.ts
```diff
@@ -246,6 +248,36 @@ class GameStore extends VuexModule implements GameState {
...
+
+    @Mutation
+    removeMarker(data: { marker: string; sync: boolean }): void {
+        this.markers = this.markers.filter(m => m !== data.marker);
+        if (data.sync) socket.emit("Marker.Remove", data.marker);
+    }
+
+    @Mutation
+    jumpToMarker(marker: string): void {
+        shape.uuid = layerManager.UUIDMap.get(marker);
```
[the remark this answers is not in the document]
LDeeJay1969: Must have been a copy/paste error. Not what I had originally. Now changed.

client/src/game/ui/selection/shapecontext.vue
```diff
@@ -77,6 +78,7 @@ import { Layer } from "../../layers/layer";
     computed: {
         ...mapState("game", ["activeFloorIndex", "assets", "locations", "notes"]),
         ...mapMutations("game", ["selectFloor"]),
+        ...mapState("game", ["locations", "assets", "notes", "markers"])
```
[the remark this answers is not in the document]
LDeeJay1969: This occurred after a few updates. I'll recommit to the most current version (same goes for changelog).

server/models/campaign.py
LDeeJay1969: Nope. I started originally with a copy of Note because it looked similar to what I wanted. After a lot of messing about with a lot a files, the result is completely different :-) 
I've changed it to reference location instead of room.

server/save.py
LDeeJay1969: As with changelog and menu.vue, I missed a few updates. Will redo this on the latest dev (probably my lack of experience with github as well ;-)

client/src/game/ui/selection/shapecontext.vue
```diff
@@ -50,6 +50,7 @@
         <li @click="addInitiative">{{ getInitiativeWord() }} initiative</li>
         <li @click="deleteSelection">Delete shapes</li>
         <li v-if="hasSingleShape()" @click="openEditDialog">Show properties</li>
+        <li @click="setMarker">Set marker</li>
```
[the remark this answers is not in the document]
LDeeJay1969: Good point. Didn't think of that. Changed now.

LDeeJay1969:
Ok, think I'm done now (again). I see at the top of this page: 

> commits into Kruptein:dev from LDeeJay1969:feature/GH-67-markers

but aparently I'm not seeing the latest? Because save.py and changelog.md are still 'old' ones. How do I rectify that?

client/src/game/ui/menu/menu.vue
```diff
@@ -196,6 +213,22 @@ export default class MenuBar extends Vue {
...
+
+    jumpToMarker(marker: string): void {
+        gameStore.jumpToMarker(marker);
+    }
+
+    nameMarker(marker: string): string {
+        if (layerManager.UUIDMap.get(marker) != undefined) {
+            return layerManager.UUIDMap.get(marker)!.name;
+        } else {
+            return "";
```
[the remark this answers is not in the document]
LDeeJay1969: This should not occur anymore since there's either a marker or there isn't. I can't think of a situation where you get a return but not know what. The undefined check is more to satisfy TS, I'm certain you won't get undefined back from the server.

server/api/socket/location.py
LDeeJay1969: Now taken care of.

Kruptein:
if you merge dev into your branch, did you resolve all the merge conflicts git throws at you ?

LDeeJay1969:
I did now. I'm still confused about CHANGELOG.MD though. It wanted me to change lines that weren't mine....? I hope I got that right. I didn't get any conflicts about save.py and menu.vue. Not sure how that worked out then...

client/src/game/ui/menu/menu.vue
```diff
@@ -196,6 +213,22 @@ export default class MenuBar extends Vue {
...
+        gameStore.removeMarker({ marker: marker, sync: true });
+    }
+
+    jumpToMarker(marker: string): void {
+        gameStore.jumpToMarker(marker);
+    }
+
+    nameMarker(marker: string): string {
+        if (layerManager.UUIDMap.get(marker) != undefined) {
+            return layerManager.UUIDMap.get(marker)!.name;
```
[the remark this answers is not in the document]
LDeeJay1969: Done that

CHANGELOG.md
```diff
@@ -43,6 +45,8 @@ All notable changes to this project will be documented in this file.
 -   Floors
     -   Create/Delete floors that are rendered on top of eachother to increase immersion
     -   Use Page Up/Down as a quick keybinding to move between floors
+-  Touch Gestures [ZachMyers3]
+-  Easier client traversing by removing _load route [ZachMyers3]
```
Kruptein: Yeah, something went wrong during your merge conflict :D

These lines are already there below, so just remove them again.

client/src/game/store.ts
```diff
@@ -244,6 +246,32 @@ class GameStore extends VuexModule implements GameState {
...
+
+    @Mutation
+    removeMarker(data: { marker: string; sync: boolean }): void {
+        this.markers = this.markers.filter(m => m !== data.marker);
+        if (data.sync) socket.emit("Marker.Remove", data.marker);
+    }
+
+    @Mutation
+    jumpToMarker(marker: string): void {
+        const shape = layerManager.UUIDMap.get(marker);
```
Kruptein: You should handle the case where the shape does not exist.

client/src/game/ui/menu/menu.vue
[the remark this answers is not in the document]
Kruptein: This is not what I meant :D 

```typescript
const shape = layerManager.UUIDMap.get(marker);
if (shape !== undefined) {
    return shape.name;
}
```
with the abvoe you only do the .get on UUIDmap once and guarantee type safeness inside the if check without an extra ! needed.

client/src/game/ui/menu/menu.vue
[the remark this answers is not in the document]
Kruptein: That's fine and good to hear that it shouldn't occur in normal situations.

I would still prefer the [?] to be returned here instead of in the template though.

client/src/game/ui/selection/shapecontext.vue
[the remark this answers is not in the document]
Kruptein: This seems to still be wrong?

server/save.py
Kruptein: This also seems to be still wrong

client/src/game/ui/menu/menu.vue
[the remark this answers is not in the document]
LDeeJay1969: I don't get it..... that's more or less what I had but it didn't show up on github...?

server/save.py
LDeeJay1969: Yes, as said for some reason I didn't see those new updates. Now updated.

client/src/game/store.ts
LDeeJay1969: Not sure what you mean here.... in which function?

client/src/game/ui/selection/shapecontext.vue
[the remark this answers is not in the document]
LDeeJay1969: Should be solved now.

CHANGELOG.md
LDeeJay1969: Sorted.

client/src/game/ui/menu/menu.vue
```diff
@@ -223,8 +223,9 @@ export default class MenuBar extends Vue {
     }
 
     nameMarker(marker: string): string {
-        if (layerManager.UUIDMap.get(marker) !== undefined) {
-            return layerManager.UUIDMap.get(marker)!.name;
+        const mrk = layerManager.UUIDMap.get(marker)!.name;
+        if (mrk !== undefined) {
+            return mrk;
```
Kruptein:
```suggestion
        const shape = layerManager.UUIDMap.get(marker);
        if (shape !== undefined) {
            return shape.name;
```
I don't know why you didn't just use the suggestion I proposed. Your current approach will still error on that first line if the returned value is undefined.

server/save.py
```diff
@@ -311,6 +308,13 @@ def upgrade(version):
             )
         db.foreign_keys = True
         Constants.get().update(save_version=Constants.save_version + 1).execute()
+    elif version == 21:
+        db.foreign_keys = False
+        db.execute_sql(
+            'CREATE TABLE IF NOT EXIST "marker" ("id" INTEGER NOT NULL PRIMARY KEY, `uuid` TEXT NOT NULL , `user_id` INTEGER NOT NULL, `room_id` INTEGER NOT NULL, FOREIGN KEY(`uuid`) REFERENCES `shape`(`uuid`) ON DELETE CASCADE, FOREIGN KEY(`room_id`) REFERENCES `room`(`id`) ON DELETE CASCADE, FOREIGN KEY(`user_id`) REFERENCES `user`(`id`) ON DELETE CASCADE)'
```
Kruptein: This SQL line still references room instead of location

client/src/game/ui/menu/menu.vue
LDeeJay1969: This was the only way I could get it working..... Your suggestion gives me an error: 
> Function lacks ending return statement and return type does not include 'unde
> fined'.

server/save.py
LDeeJay1969: I hate these local/remote versions... ;-)
Copied from the wrong version. Apologies. Now corrected.

client/src/game/ui/menu/menu.vue
Kruptein: It's only the start that needs to be replaced that I posted.
The entire function is

```typescript
nameMarker(marker: string): string {
        const shape = layerManager.UUIDMap.get(marker);
        if (shape !== undefined) {
            return shape.name;
        } else {
            return "";
        }
    }
```

LDeeJay1969: Probably my lack of TS-knowledge. I still don't get the difference but hey... it works! ;-)
I'll try to dig up some TS course. It's not really JS... ;-)

Kruptein: your first line calls the UUIDMap.get function which can return either a value or undefined.  Then you call .name on that result.  If you do it straight away TS will warn you that this can result in an error if get returns undefined.  You then added a ! to tell TS that it does not need to worry and you're sure you know what you're doing.
Thing is that that is bound to go wrong at one point and thus better avoided.  Here avoiding it is relatively easy by calling .get() and then using an if check to verify whether it is undefined or not.

client/src/game/store.ts
LDeeJay1969: Got it. Updated now.

server/save.py
```diff
@@ -309,6 +308,13 @@ def upgrade(version):
             )
         db.foreign_keys = True
         Constants.get().update(save_version=Constants.save_version + 1).execute()
+    elif version == 21:
+        db.foreign_keys = False
+        db.execute_sql(
+            'CREATE TABLE IF NOT EXIST "marker" ("id" INTEGER NOT NULL PRIMARY KEY, `uuid` TEXT NOT NULL , `user_id` INTEGER NOT NULL, `location_id` INTEGER NOT NULL, FOREIGN KEY(`uuid`) REFERENCES `shape`(`uuid`) ON DELETE CASCADE, FOREIGN KEY ("location_id") REFERENCES "location" ("id") ON DELETE CASCADE, FOREIGN KEY(`user_id`) REFERENCES `user`(`id`) ON DELETE CASCADE)'
```
Kruptein:
```suggestion
            'CREATE TABLE IF NOT EXISTS "marker" ("id" INTEGER NOT NULL PRIMARY KEY, `uuid_id` TEXT NOT NULL, `user_id` INTEGER NOT NULL, `location_id` INTEGER NOT NULL, FOREIGN KEY(`uuid_id`) REFERENCES `shape`(`uuid`) ON DELETE CASCADE, FOREIGN KEY ("location_id") REFERENCES "location" ("id") ON DELETE CASCADE, FOREIGN KEY(`user_id`) REFERENCES `user`(`id`) ON DELETE CASCADE)'
```

Kruptein: There was a typo in the exists keyword and the uuid field was missing its _id postfix

server/models/campaign.py
```diff
@@ -166,6 +167,18 @@ def as_dict(self):
...
 
+class Marker(BaseModel):
+    from .shape import Shape
+
+    uuid = ForeignKeyField(Shape, backref="markers", on_delete="CASCADE")
+    user = ForeignKeyField(User, backref="markers", on_delete="CASCADE")
+    location = ForeignKeyField(Location, backref="markers", on_delete="CASCADE")
+
+    def __repr__(self):
+        return f"<Marker {self.shape.uuid} {self.location.get_path()} - {self.user.name}"
```
Kruptein: self.shape does not exist.  to fix it it should be self.uuid.uuid, which is kinda awkward.  So I suggest that you rename the uuid field to shape (don't forget to change the save.py migration query and marker.py socket file if you do this)

client/src/game/ui/menu/menu.vue
```diff
@@ -324,14 +358,18 @@ DIRECTORY.CSS changes
 .menu-accordion-subpanel > * {
     padding: 5px;
     display: flex;
-    justify-content: space-evenly;
     align-items: center;
+    text-align: left;
```
Kruptein: I would apply this change only on the markers part for now.

Some redesign of the side bar is needed in the future, but having all the other elements also left-aligned is kinda weird imo.

(same thing applies to the removal of the justify-content)

client/src/game/ui/selection/shapecontext.vue
```diff
@@ -50,6 +50,7 @@
         <li @click="addInitiative">{{ getInitiativeWord() }} initiative</li>
         <li @click="deleteSelection">Delete shapes</li>
         <li v-if="hasSingleShape()" @click="openEditDialog">Show properties</li>
+        <li v-if="hasSingleShape()" @click="setMarker">Set marker</li>
```
Kruptein: After playing with it, we might want to hide this if it already is a marker or maybe replace it with a remove marker element.

server/save.py
LDeeJay1969: Changed

server/models/campaign.py
LDeeJay1969: Changed.

client/src/game/ui/selection/shapecontext.vue
LDeeJay1969: Right, I've been trying this. vue side is ok, gamestore side is ok. I can query the markers array in the gamestore from vue but I can't get the result back to vue, it's always undefined. Do I need to export something from gamestore or do I need to query the db directly from vue?

Kruptein: You could add a method to the gamestore, but you do have everything you need in this file.  You import the markers from the store in your computed section and you have the access to the shape and thus it's id.

client/src/game/ui/menu/menu.vue
LDeeJay1969: Done.

LDeeJay1969:
Do you still want the marker removed from campaign.py?  I've changed the location so that Layer is accessed before Marker and that resolved the error (in my case).

Kruptein:
It's not a major show stopper, but I do in general like the imports to be sorted if possible, but the primary reason is that the campaign.py file is becoming pretty large.  So splitting it up wouldn't hurt I think.

server/api/socket/marker.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+@sio.on("Marker.Remove", namespace="/planarally")
+@auth.login_required(app, sio)
+async def delete_marker(sid, uuid):
+    sid_data = state.sid_map[sid]
+    user = sid_data["user"]
+    location = sid_data["location"]
+
+    marker = Marker.get_or_none(uuid=uuid)
```
Kruptein, line 32:
```suggestion
    marker = Marker.get_or_none(shape=uuid, user=user)
```
the `uuid` field no longer exists and additionally you need to explicitly check the user, or you could accidentally remove someone else's marker of the same shape.

server/api/socket/marker.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@sio.on("Marker.New", namespace="/planarally")
+@auth.login_required(app, sio)
+async def new_marker(sid, data):
+    sid_data = state.sid_map[sid]
+    user = sid_data["user"]
+    location = sid_data["location"]
+
+    marker = Marker.get_or_none(shape=data)
```
Kruptein, line 14:
```suggestion
    marker = Marker.get_or_none(shape=data, user=user)
```

client/src/game/ui/menu/menu.vue
```diff
@@ -68,6 +68,21 @@
                                 <div v-if="!notes.length">No notes</div>
                             </div>
                         </div>
+                        <!-- MARKERS -->
```
Kruptein, line 71: I would move the markers section out of the DM only section, I can see some players also wanting to jump around between markers in some circumstances.

Kruptein:
I also changed something else that affects the save format, so you'll need to pull some changes and readjust your save.py file.  I know it's a bit clunky, so sorry for the inconvenience

LDeeJay1969:
I hope I got the save.py concept right. I don't undestand though, why under version 20 Booleanfield is mentioned twice and under 21 Integerfield is mentioned as well (as is Booleanfield twice).

LDeeJay1969:
What I also don't get is how my own gh-rep will keep up-2-date with dev. It's always conflicting with CHANGELOG, now also with save.py.  When I look at dev directly, I see completely different versions. I thought everything in dev would be reflected in my fork?

Kruptein:
> I hope I got the save.py concept right. I don't undestand though, why under version 20 Booleanfield is mentioned twice and under 21 Integerfield is mentioned as well (as is Booleanfield twice).

Haha that' actually just a stupid error on my end :p there is no need for it indeed.

Kruptein:
> What I also don't get is how my own gh-rep will keep up-2-date with dev. It's always conflicting with CHANGELOG, now also with save.py. When I look at dev directly, I see completely different versions. I thought everything in dev would be reflected in my fork?

Are you actively pulling in my dev into your dev ?

server/api/socket/marker.py
LDeeJay1969: Changed

client/src/game/ui/menu/menu.vue
LDeeJay1969: Done.

server/api/socket/marker.py
LDeeJay1969: Done

server/api/socket/marker.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+@sio.on("Marker.Remove", namespace="/planarally")
+@auth.login_required(app, sio)
+async def delete_marker(sid, uuid):
+    sid_data = state.sid_map[sid]
+    user = sid_data["user"]
+    location = sid_data["location"]
+
+    marker = Marker.get_or_none(shape_id=uuid)
```
Kruptein, line 32: This one is not updated yet ?

LDeeJay1969: Yes, I did.... I noticed that `marker = Marker.get_or_none(uuid=uuid)` hadn't changed yet. Since the field name in the table is shape_id I changed it.

Kruptein: No I mean it's missing user=user

LDeeJay1969: My bad, I changed the add-function instead of the delete-function.... sorted now.

Kruptein: 😄 It needs to be there in both cases

Kruptein: otherwise other users won't be able to create a marker for a similar shape

LDeeJay1969: I know I shouldn't do two things at the same time.... 
Focus Leon, FOCUS!!!  ;-)

LDeeJay1969:
Yay !! Apologies for the mess. It was a steep learning curve (still ongoing) for someone with no knowledge of TS and VUE but I've learned a lot so thanks for the input and the patience. 

I've got some general questions (PA related but not directly functional). Where do you want me to pose those?

Kruptein:
Hey no problems at all, I'm not gonna claim this to be the easiest or cleanest code base by any means, so some struggles are very normal.  If you happen to have discord, I created a server yesterday, https://discord.gg/mubGnTe.   Otherwise feel free to mail me or send me something on reddit :)

Issue #199, opened 2020-01-15
Use router.replace to forget _load


Kruptein:
I currently use a _load temporary route to show while prefetching stuff.  This ends up in the window history.

This can be remedied by using router replace.

ZachMyers3:
What I've done in Vue projects is keep a loading flag and nest my component's ```<div>``` within a ```v-if```. I have a simple example below.

You could potentially have a loading state in the store, and then do something similar?

Just throwing this out there. This has allowed me to lazy load some components fairly easily with seamless loading built in.

```vue
<template>
    <div v-if="!loading">
        <!-- component template here -->
    </div>
    <div v-else>
        <!-- circular loading overlay -->
        <v-overlay>
            <v-progress-circular
                indeterminate
                size="64"
            ></v-progress-circular>
        </v-overlay>
    </div>
</template>
```

Kruptein:
This might indeed be a more elegant approach

ZachMyers3:
Trying this locally, trying to get loading to be based for ```coreStore.initialized```. I can get routing and authentication to work properly with my router adjustments but the game map isn't loading. Curious if you might know more. New router is below.

```typescript
// route.ts
router.beforeEach(async (to, _from, next) => {
    if (to.matched.some(record => record.meta.auth)) {
        // authentication is required, but the store isnt authenticated
        if (!coreStore.initialized) {
            console.log("Not authenticated");
            // verify with the api to see if really authenticated
            const response = await fetch("/api/auth");
            if (response.ok) {
                const data = await response.json();
                // api reports authentication, adjust state and proceed
                coreStore.setAuthenticated(true);
                coreStore.setUsername(data.username);
                coreStore.setEmail(data.email);
                coreStore.setInitialized(true);
                next();
            } else {
                // truly not authenticated, redirect to login
                next({
                    path: "/auth/login",
                    query: {
                        redirect: to.path,
                    },
                });
            }
        } else {
            // the store says client is initialized
            next();
        }
    } else {
        // no authentication required
        next();
    }
});
```

Kruptein:
Just at a glance there are some checks that are dropped from the original code.
You no longer check the `coreStore.authenticated` field for the auth records. As soon as the coreStore is initialized the if check will be skipped entirely and no auth is verified.
The reason this currently works is because the data.auth check is also skipped upon receiving data from the `/api/auth` call.  The returned response can be 200 but with a failed authentication.

[...]

PR #215, opened 2020-02-21
Router

ZachMyers3:
For #199 

I was able to remove the _load directory from the router and move the concept into a loading state in the core store. Now on router.beforeEach() this loading state gets enabled and then router.afterEach() disables the state. From there I was able to watch the state from within App.vue and use and to display LoadComponent given the state.


ZachMyers3:
I added a component I found [here](https://github.com/ankurk91/vue-loading-overlay).

Essentially just a nice looking wheel overlay instead of the ```Loading...``` text before. 

If you don't like the additional dependencies with the loading overlay I can remove it.

On top of this anything (more than just authentication) can use this loading overlay by just adjusting the Vuex loading state on the core store.

client/src/router.ts
```diff
@@ -78,25 +72,51 @@ export const router = new Router({
...
+                    });
+                }
+            } else {
+                // truly not authenticated, redirect to login
+                next({
+                    path: "/auth/login",
+                    query: {
+                        redirect: to.path,
+                    },
+                });
```
Kruptein: I feel like this currently is just a more verbose way to write the original code.  What do you want to accomplish here that couldn't be accomplished with the old code and just dropping the `/_load` if case and next() call?

ZachMyers3: So this is a little more complex, but I feel it gives a little more flexibility.

The old logic was making the API call on every directory wasn't _load if you weren't initialized. Now on a refresh event this will only make that call on a ```meta.auth``` path.

I tried a couple different things, actually tried to push all the API/authentication login within the coreStore but couldn't get things to wait properly for some variables in the router to populate to do it.

I also broke out my login redirect ```next()``` calls, which make it appear a little bigger. I could collapse those if you'd like.

client/src/App.vue
```diff
@@ -1,9 +1,30 @@
...
+<script lang="ts">
+import { Component, Vue } from "vue-property-decorator";
+import LoadComponent from "@/core/components/load.vue";
+import Loading from "vue-loading-overlay";
+import "vue-loading-overlay/dist/vue-loading.css";
+import { coreStore } from "@/core/store";
+
+@Component({
+    components: {
+        LoadComponent,
```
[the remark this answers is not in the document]
ZachMyers3: It sure is, I'll get it removed on my next commit

ZachMyers3:
Another question I had, just trying to play around with the loading state. For the life of me I couldn't figure out where the core map draw logic is. I just wanted to have ```game.vue``` display the loading wheel while the field rendered but couldn't determine where to turn loading on/off. This isn't exactly related to this PR, but I was curious.

client/src/router.ts
Kruptein: In practice the old code would only do the auth call on the very first route enter though.  As the store will be initialized after this first visit.

Kruptein:
The core draw loop happens in `src/game/layers/manager` the drawLoop function

client/src/router.ts
ZachMyers3: Yeah exactly, all this really does different is initialize the store on the first time it's needed.

If you go line by line the new implementation is actually less lines if you format the next reroutes the same way, so I guess I'm confused where the verbosity is at.

Kruptein: I must be missing something here then.  This is the old code without the _load, but using the new setLoading.  It's 22 lines 
```
router.beforeEach(async (to, from, next) => {
    coreStore.setLoading(true);
    if (!coreStore.initialized) {
        const response = await fetch("/api/auth");
        if (response.ok) {
            const data = await response.json();
            if (data.auth) {
                coreStore.setAuthenticated(true);
                coreStore.setUsername(data.username);
                coreStore.setEmail(data.email);
            }
            coreStore.setInitialized(true);
            router.push(to.path);
        } else {
            console.error("Authentication check could not be fulfilled.");
        }
    } else if (to.matched.some(record => record.meta.auth) && !coreStore.authenticated) {
        next({ path: "/auth/login", query: { redirect: to.path } });
    } else {
        next();
    }
});
```

and this is your suggestion with all comments stripped out and the next() arguments all on one line.  It's 26 lines and the second half contains a lot of } else {'s that require slightly more cognitive load to see what actually is happening.
```
router.beforeEach(async (to, _from, next) => {
    coreStore.setLoading(true);
    if (to.matched.some(record => record.meta.auth)) {
        if (!coreStore.authenticated) {
            const response = await fetch("/api/auth");
            if (response.ok) {
                const data = await response.json();
                if (data.auth) {
                    coreStore.setAuthenticated(true);
                    coreStore.setUsername(data.username);
                    coreStore.setEmail(data.email);
                    coreStore.setInitialized(true);
                    next();
                } else {
                    next({path: "/auth/login",query: {redirect: to.path,},});
                }
            } else {
                next({path: "/auth/login",query: {redirect: to.path,},});
            }
        } else {
            next();
        }
    } else {
        next();
    }
});
```

I also just feel that the original code is easier to follow.  First there is a check if the store has been initialized if not, perform the init steps.  After that there is a check to see if you're accessing a auth restricted page whether you are indeed authenticated.  and if the last else is reached you know everything has been resolved.

The current proposal also never sets the coreStore.initialized state until a auth route is visited. I'm aware there is nothing except the router that currently uses the initialized state, but it could very well be used for something in the future.

I hope I'm not coming over to aggressive here as I honestly appreciate the input, I'm just trying to get most out of it ;)

ZachMyers3: What do you think of this?

```typescript
router.beforeEach(async (to, _from, next) => {
    coreStore.setLoading(true);
    if (to.matched.some(record => record.meta.auth)) {
        // authentication is required, but the store isnt authenticated
        if (!coreStore.initialized) {
            const response = await fetch("/api/auth");
            if (response.ok) {
                const data = await response.json();
                if (data.auth) {
                    coreStore.setAuthenticated(true);
                    coreStore.setUsername(data.username);
                    coreStore.setEmail(data.email);
                }
                coreStore.setInitialized(true);
                router.push(to.path);
            } else {
                console.error("Authentication check could not be fulfilled.");
            }
        } else {
            next();
        }
    } else {
        // no authentication required
        next();
    }
});
```

I get where you're coming from, the first one does appear simpler.

This takes your core initialization logic from the old code and removes any rerouting to _load. I think this is better than my previous one.

Just curious, if ```.itialized``` and ```.authenticated``` only ever get turned off/on in the same instances, what's the point of having both? You're not coming off aggressive, and I appreciate the input 8)

ZachMyers3:
Here's what has given me the best success on loading for the game.

```typescript
// /game/api/events.ts
socket.on("connect", () => {
    console.log("Connected");
    coreStore.setLoading(false);
});
```

```typescript
// game/game.vue
@Component({
    components: {
        "tool-bar": Tools,
        "selection-info": SelectionInfo,
        "prompt-dialog": Prompt,
        "confirm-dialog": ConfirmDialog,
        "floor-select": FloorSelect,
        "menu-bar": MenuBar,
        "initiative-dialog": Initiative,
        "zoom-slider": vueSlider,
        "note-dialog": NoteDialog,
        "label-dialog": LabelManager,
        "dm-settings": DmSettings,
    },
    beforeRouteEnter(to, from, next) {
        coreStore.setLoading(true);
        createConnection(to);
        next();
    },
    beforeRouteLeave(to, from, next) {
        socket.disconnect();
        next();
    },
})
export default class Game extends Vue { }
```

It seems like the biggest slowdown on the loading is actually getting connected to the socket. I played around with turning the loading flag off after the drawloop but on my local machine that runs fairly fast.

Curious on your thoughts of this. It works decently well and provides more input than it previously did.

client/src/router.ts
ZachMyers3: I've got these changes in my latest commit, let me know what you think

client/src/App.vue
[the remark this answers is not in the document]
ZachMyers3: In latest commit.

client/src/core/components/load.vue
```diff
@@ -9,5 +9,9 @@ import Component from "vue-class-component";
 @Component
 export default class LoadComponent extends Vue {
```
Kruptein: I think this component can be removed now entirely

client/src/dashboard/main.vue
```diff
@@ -98,6 +98,14 @@ export default class Dashboard extends Vue {
...
         }
     }
+
+    beforeCreate(): void {
+        console.log("Before create");
+    }
+
+    created(): void {
+        console.log("Created");
+    }
```
Kruptein: These should obviously go ;)

client/src/router.ts
```diff
@@ -78,25 +72,33 @@ export const router = new Router({
...
-    if (!coreStore.initialized && to.path !== "/_load") {
-        next({ path: "/_load" });
-        const response = await fetch("/api/auth");
-        if (response.ok) {
-            const data = await response.json();
-            if (data.auth) {
-                coreStore.setAuthenticated(true);
-                coreStore.setUsername(data.username);
-                coreStore.setEmail(data.email);
+router.beforeEach(async (to, _from, next) => {
```
Kruptein: You never check the .authenticated field on auth records now, nor do you redirect people to /login.

I'm gonna keep reiterating this :p what is wrong with the old code and just changing the _load stuff to the new setLoading stuff ?

ZachMyers3: Sorry, I (finally) get what you're saying now.

My latest commit has these changes.

client/src/dashboard/main.vue
ZachMyers3: Gone it latest commit.

client/src/core/components/load.vue
ZachMyers3: I've deleted the entire load.vue file in the latest commit.

ZachMyers3:
I think this resolves everything we've been talking about.

Issue #216, opened 2020-02-23
Display Version Info on Client 


ZachMyers3:
A nice QoL improvement to see how in/out of date your server is.

ZachMyers3:
Could this be pulled from ```packages.json```?

I see the package has a version tag, it'd just need to be changed on version upgrades. This is only accounting for the client version though. How would we approach getting the server version, and API endpoint?

```json
// packages.json
{
  "name": "planarally-client",
  "version": "0.18.2",
  "description": "A companion tool for when you travel into the planes.",
  "scripts": {
    "serve": "vue-cli-service serve",
    "build": "npm run clean && vue-cli-service build --no-clean --modern",
    "lint": "vue-cli-service lint",
    "clean": "rimraf ../server/static/js/ && rimraf ../server/static/css/"
  }
}
```

Kruptein:
client has it in it's package.json and package-lock, server has it in the VERSION file.

Although it could change in the future, you can assume for now that client and server are always on the same versions.

ZachMyers3:
Would the best bet be to implement an API endpoint for the server to pull into the client? I'm not too fond of the idea of pulling from the server directory from the client.

Kruptein:
Sure, I believe you could also parse it from the package json itself, but that involves some potential security issues, so just having a server api endpoint is absolutely fine.

PR #220, opened 2020-02-25
Display version on client

ZachMyers3:
For #216.

Just a simple API endpoint from the server. Currently pulling and displaying on the login prompt's legend. I figured there were other places better suited for this but wanted input before throwing them around.

Let me know what you think!

### DIFF
--- a/client/src/router.ts
+++ b/client/src/router.ts
@@ -75,15 +75,17 @@ export const router = new Router({
 router.beforeEach(async (to, _from, next) => {
     coreStore.setLoading(true);
     if (!coreStore.initialized) {
-        const response = await fetch("/api/auth");
-        if (response.ok) {
-            const data = await response.json();
-            if (data.auth) {
+        const authResponse = await fetch("/api/auth");
+        const versionResponse = await fetch("/api/version");
+        if (authResponse.ok) {
+            const authData = await authResponse.json();
+            const versionData = await versionResponse.json();
+            if (authData.auth) {
                 coreStore.setAuthenticated(true);
-                coreStore.setUsername(data.username);
-                coreStore.setEmail(data.email);
-                coreStore.setVersion(data.version);
+                coreStore.setUsername(authData.username);
+                coreStore.setEmail(authData.email);
             }
+            coreStore.setVersion(versionData.version);
             coreStore.setInitialized(true);
             router.push(to.path);
         } else {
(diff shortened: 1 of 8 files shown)


ZachMyers3:
Here's a screenshot of the display.

image

server/api/http/version.py
```diff
@@ -0,0 +1,14 @@
+from aiohttp import web
+
+async def get_version(request: web.Request):
+    try:
+        with open('VERSION', 'r') as version_file:
+            version_data = version_file.read()
+    except:
+        version_data = "0.0.0"
```
Kruptein: I would just return a 500 error because if this happens, there is something wrong with the server.

ZachMyers3:
If we want to inject it into the initialization state, should we tie it into the authentication query? Otherwise we'd await two responses on initialization.

Kruptein:
I get your point, but I think that from an API standpoint, returning version info in the is_authed endpoint is strange.

you can use Promise.all to start two parallel request that won't block on eachother.

ZachMyers3:
I definitely agree with you, I think this is the cleaner way to do it. 

Latest commit has it this way.

client/src/router.ts
```diff
@@ -75,14 +75,17 @@ export const router = new Router({
 router.beforeEach(async (to, _from, next) => {
     coreStore.setLoading(true);
     if (!coreStore.initialized) {
-        const response = await fetch("/api/auth");
-        if (response.ok) {
-            const data = await response.json();
-            if (data.auth) {
+        const authResponse = await fetch("/api/auth");
+        const versionResponse = await fetch("/api/version");
+        if (authResponse.ok && versionResponse.ok) {
```
Kruptein: Now you block the second request by waiting on the first.  What you want to do is something akin to

```
const [authResponse, versionResponse] = await Promise.all([fetch("/api/auth"), fetch("/api/version"]);
coreStore.setVersion(versionResponse.version);
if (authResponse.auth) { ... }
```

server/api/http/version.py
```diff
@@ -0,0 +1,14 @@
+from aiohttp import web
+
+async def get_version(request: web.Request):
+    try:
+        with open('VERSION', 'r') as version_file:
+            version_data = version_file.read()
+    except:
+        return web.HTTPError()
```
Kruptein:
```suggestion
        return web.HTTPInternalServerError(reason="Version file could not be loaded")
```

client/src/router.ts
ZachMyers3: Latest commit has these changes.

I've actually never done this before, so it's new!

ZachMyers3:
Works for me! Thanks!

Kruptein:
I did just notice that somehow, the package-lock is empty in dev.  Could you reinstantiate that file? :D

Kruptein:
On second thoughts, I'll just update it on dev immediately

Issue #235, opened 2020-03-12
Being able to add a number label to tokens 


adunis:
I'm in need of a feature that IMHO it's very necessary. When dealing with a large number of tokens using the same image, even if you name token with a different number it becomes too hard to differentiate between Goblin 1, 2, 3 etc. 

I suggest being able to add a 'label' to a token that is overlayed over the token image. 

I made a quick sketch of  how it could look 

<img width="494" alt="NumberedGoblins" src="https://user-images.githubusercontent.com/47754845/76576541-a9988700-64c2-11ea-9b58-8462d1d4ad31.png">

Kruptein:
This should be relatively easy to add.  I do have some questions in regards to desired behaviour.

* Would this be something you explicitly enable for a certain token or rather be opt out per token? Or even have this as a user setting?
* Do you expect some sort of auto-numbering?  If so
   * Do you expect this to be limited to copies of existing tokens on the board?  (e.g. say you have a particular monster token which you assign '1' and you copy/paste the token, should it automatically receive 2 as its number?)
   * Do you expect the numbering to be across all tokens instead? Green monster 1, Blue monster 2, Green monster 3 ...
    * What if I remove a token with number 2 when there is 1,2,3 and I create a new token, should it become 4 or 2 ?

adunis:
> Do you expect this to be limited to copies of existing tokens on the board? (e.g. say you have a particular monster token which you assign '1' and you copy/paste the token, should it automatically receive 2 as its number?)

Yes, this feature would speed up setting up a combat. 

> Do you expect the numbering to be across all tokens instead? Green monster 1, Blue monster 2, Green monster 3 ..

No, only to the tokens I want labelled with a number. Either by manually editing each token, or by copy-pasting an already labelled token. I except having a single non-labelled token on the map, surrounded by multiple copies of a single token with different number labels. 

> What if I remove a token with number 2 when there is 1,2,3 and I create a new token, should it become 4 or 2 

When I copy one of those tokens (1 or 2 or 3) and paste it, I except the new token copy to be labelled as a 4. When I drag and drop the token from the asset menu, I except the new token to not be labelled.

Kruptein:
Alright that's clear, I'll see what I can do this weekend :)

Kruptein:
Some more questions:

[...]

PR #239, opened 2020-03-17
Shape badges

Kruptein:
This PR starts work on issue #235 . It adds the notion of shape badges that show a token number with auto increment on copy paste.

### DIFF
--- a/client/src/game/comm/types/shapes.ts
+++ b/client/src/game/comm/types/shapes.ts
@@ -20,7 +20,7 @@ export interface ServerShape {
     is_token: boolean;
     options?: string;
     badge: number;
-    showBadge: boolean;
+    show_badge: boolean;
 }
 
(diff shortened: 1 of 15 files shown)


adunis:
Thanks for adding this feature! 

I tested briefly locally and there are three simple things to fix:

- When I add an asset as a token, and when I click ' Show Badge', I except that token border color (which defines the badge background color) to be 'white' instead of 'black with transparency 100%' 
- When I copy paste a token with 'Show Badge' checked, I except that also the token border color to be copied. 
- When I reload the session, the border color of a token with ' Show Badge ' checked gets reset to 'black with transparency 100%'. This happens only to tokens created with assets, not to shapes.

Kruptein:
I'm going to make seperate tickets for this, but good feedback, I only tested with quickly created tokens instead of actual assets so forgot to check this.

PR #284, opened 2020-04-07
Add tiered config file

ASMfreaK:
This adds a secondary but more important config file in data dir. Any option set in that file override settings from previous files. This file can contain only a handful of options. All other will be taken from base config file.


Kruptein:
I would just like you to add something to the changelog file, otherwise this looks simple but effective.

ASMfreaK:
Will this suffice?

PR #287, opened 2020-04-09
Documentation update

jirsat:
While setting up my own instance (on RaspberryPi) I found some missing info in documentation. This PR should update installation instruction and help newcomers to selfhost.

Changes:
Added example configuration for nginx.
Added example systemd service file to run server in background.
Mentioned above files in documentation.
Clarified installation instruction (building client)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -142,6 +142,7 @@ Additionally you'll probably need to modify your router to allow access from
 the outside world on your personal computer on port 8000.
 Look up information on `port forwarding` for more info on this topic.
 
+In folder "extra", you can find example systemd.service and nginx server configuration.
 
 .. _python:
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -119,11 +119,14 @@ Make sure you have a valid python 3.6+ installation.
 See :ref:`python` if you don't know how.
 
 1. Download the latest source code from `here <https://github.com/Kruptein/PlanarAlly/releases/>`_.
-2. Open the server folder.
-3. Execute `pip install -r requirements.txt`.
-4. Run the server by executing `python planarserver.py`.
-5. Optionally you can configure the server in the `server_config.cfg` file.
-6. Restart the server after applying a change.
+2. Open the client folder.
+3. Execute `npm install`.
+4. Execute `npm run build`.
+5. Return to the PlanarAlly folder and open the server folder.
+6. Execute `pip install -r requirements.txt`.
+7. Run the server by executing `python planarserver.py`.
+8. Optionally you can configure the server in the `server_config.cfg` file.
+9. Restart the server after applying a change.
 
 Configuring
 ************
--- a/extra/nginx
+++ b/extra/nginx
@@ -1,0 +1,72 @@
+#Using unix socket as upstream server
+upstream pa {
+	server unix:/tmp/planarally.sock fail_timeout=0;
+}
+
+#Forward http to https
+server {
+    listen 80;
+    listen [::]:80; #IPV6
+    server_name dnd.example.com;
+    return 301 http://www.$server_name$request_uri; 
+}
+
+
+server {
+	listen 443 ssl http2;
+	listen [::]:443 ssl http2; #IPV6
+
+	index index.html index.htm index.php index;
+  
+	client_max_body_size 4G;
+  server_name dnd.example.com;
+	
+	#protection against abuse - without it anyone can create their account
+	satisfy any;
+  #httpaswd for anyone outside local network
+	auth_basic	"Enter credentials";
+	auth_basic_user_file /path/to/.httpaswd;
+  #allow from local network - adjust for your network
+	allow 192.168.0.0/16; 
+	allow fe80::/10;
+
+	location / {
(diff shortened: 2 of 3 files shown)


extra/nginx
```diff
@@ -0,0 +1,72 @@
...
+upstream pa {
+	server unix:/tmp/planarally.sock fail_timeout=0;
+}
+
+#Forward http to https
+server {
+    listen 80;
+    listen [::]:80; #IPV6
+    server_name dnd.example.com;
+    return 301 http://www.$server_name$request_uri;
```
Kruptein, line 11: I believe this should be https ;)

extra/nginx
```diff
@@ -0,0 +1,72 @@
...
+    server_name dnd.example.com;
+    return 301 http://www.$server_name$request_uri; 
+}
+
+
+server {
+	listen 443 ssl http2;
+	listen [::]:443 ssl http2; #IPV6
+
+	index index.html index.htm index.php index;
```
Kruptein, line 19: Just to keep it simple a single `index.html` should suffice

extra/nginx
```diff
@@ -0,0 +1,72 @@
...
+  
+	client_max_body_size 4G;
+  server_name dnd.example.com;
```
Kruptein, line 22: This and a couple of other lines seem to be indented differently

extra/nginx
```diff
@@ -0,0 +1,72 @@
...
+	
+	#protection against abuse - without it anyone can create their account
+	satisfy any;
+  #httpaswd for anyone outside local network
+	auth_basic	"Enter credentials";
+	auth_basic_user_file /path/to/.httpaswd;
+  #allow from local network - adjust for your network
+	allow 192.168.0.0/16; 
+	allow fe80::/10;
```
Kruptein, line 31: I'm not familiar with this, but that's interesting!  It should become a server configuration at one point to disable new account creations etc.

I would maybe consider commenting this out as an example rather than as a default thing though

docs/installation.rst
```diff
@@ -119,11 +119,14 @@ Make sure you have a valid python 3.6+ installation.
...
-5. Optionally you can configure the server in the `server_config.cfg` file.
-6. Restart the server after applying a change.
+2. Open the client folder.
+3. Execute `npm install`.
+4. Execute `npm run build`.
+5. Return to the PlanarAlly folder and open the server folder.
+6. Execute `pip install -r requirements.txt`.
+7. Run the server by executing `python planarserver.py`.
+8. Optionally you can configure the server in the `server_config.cfg` file.
+9. Restart the server after applying a change.
```
Kruptein, line 129: Building the client stuff manually will no longer be mandatory starting with next release.  There will be planarally-bin.zip and tar.gz which already include these files.  So while it currently is indeed required, hopefully this will no longer be required next release.

extra/planarally.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=PlanarAlly Server
+After=network.target
+
+[Service]
+Type=simple
+#User www-data is necessary for nginx to work
+User=www-data
+WorkingDirectory=/path/to/PlanarAlly
+ExecStart=/path/to/python3.6+ /path/to/PlanarAlly/server/planarserver.py
```
ASMfreaK, line 10: We should add a shebang to planarserver.py as the first line:
```
#!/usr/bin/env python3
```

docs/installation.rst
```diff
@@ -119,11 +119,14 @@ Make sure you have a valid python 3.6+ installation.
...
-2. Open the server folder.
-3. Execute `pip install -r requirements.txt`.
-4. Run the server by executing `python planarserver.py`.
-5. Optionally you can configure the server in the `server_config.cfg` file.
-6. Restart the server after applying a change.
+2. Open the client folder.
+3. Execute `npm install`.
+4. Execute `npm run build`.
+5. Return to the PlanarAlly folder and open the server folder.
+6. Execute `pip install -r requirements.txt`.
```
ASMfreaK, line 126: We also need to instruct users to install PA in a virtualenv - this way their systems will be left intact.

ASMfreaK:
`````markdown
5. Optionally create a new virtualenv to keep your system installation intact:
```
python3 -m venv .planarally_venv
```
6. If you created a virtualenv, you need to activate it: `source .planarally_venv`
`````

File: extra/planarally.service (hunk 6 above)
Comment (ASMfreaK) [reply extra/planarally.service:10]: This should also say about virtualenvs:
```systemd
# if you installed PA in a virtualenv, comment the previous line 
# and use the following line:
# ExecStart=/path/to/virtualenv/bin/python /path/to/PlanarAlly/server/planarserver.py
```

jirsat: This is great, thanks O:-) i'll use this on my machine :D i haven't thought of this

jirsat:
ok, i'll create new PR with this in mind

extra/planarally.service
jirsat: > We should add a shebang to planarserver.py as the first line and make it executable (chmod +x `planarserver.py`):
> 
> ```
> #!/usr/bin/env python3
> ```

i agree

PR #290, opened 2020-04-10
Documentation update

jirsat:
While setting up my own instance (on RaspberryPi) I found some missing info in documentation. This PR should update installation instruction and help newcomers to selfhost. New PR after problems with #287 

Changes:
Added example configuration for nginx.
Added example systemd service file to run server in background.
Mentioned above files in documentation.
Clarified installation instruction (building client, using virtual environment)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -142,6 +142,7 @@ Additionally you'll probably need to modify your router to allow access from
 the outside world on your personal computer on port 8000.
 Look up information on `port forwarding` for more info on this topic.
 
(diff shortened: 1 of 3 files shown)


Kruptein:
Hey, I'm thankful for the PR but this is now better suited for the planarally-docs repository I think.

The self-hosting page (https://www.planarally.io/tutorial/setup/self-hosting/) could use your changes in the first file in this PR.
And your example files would also much better fit that repository.  In an examples folder or such for example.

Kruptein:
The repo can be found at https://github.com/Kruptein/planarally-docs

PR #291, opened 2020-04-11
Draft of Rotation Feature

Milamberz:
Implemented "Rotation" feature for Assets, Rects and Polygons.
Right-Click on a object and select "Rotate", rotates by 90 degrees.

### DIFF
--- a/client/src/game/shapes/asset.ts
+++ b/client/src/game/shapes/asset.ts
@@ -35,8 +35,7 @@ export class Asset extends BaseRect {
 
             if (this.rotationDegrees == 0) {
                 ctx.drawImage(this.img, g2lx(this.refPoint.x), g2ly(this.refPoint.y), g2lz(this.w), g2lz(this.h));
-            }
-            else {
+            } else {
 
                 ctx.save();
 
--- a/client/src/game/shapes/asset.ts
+++ b/client/src/game/shapes/asset.ts
@@ -1,7 +1,7 @@
 // Developer Note: Rotation
 // We only allow 0, 90 ,180 and 270 degrees
 // On an Asset, we need to keep the rotationDegrees to ensure the image is drawn correctly
-// We change the Height and Width for 90 and 270 roations so the selection, resize and bisection code still works
+// We change the Height and Width for 90 and 270 rotations so the selection, resize and bisection code still works
 
 import { InitiativeData } from "@/game/comm/types/general";
 import { ServerAsset } from "@/game/comm/types/shapes";
@@ -58,7 +58,7 @@ export class Asset extends BaseRect {
                 else
                     ctx.drawImage(this.img, g2lx(this.refPoint.x), g2ly(this.refPoint.y), g2lz(this.w), g2lz(this.h));
 
-                // we’re done with the rotating so restore the unrotated context
+                // weâ€™re done with the rotating so restore the unrotated context
                 ctx.restore();
             }
         } catch (error) {
--- a/client/src/game/shapes/asset.ts
+++ b/client/src/game/shapes/asset.ts
@@ -1,3 +1,8 @@
+// Developer Note: Rotation
+// We only allow 0, 90 ,180 and 270 degrees
+// On an Asset, we need to keep the rotationDegrees to ensure the image is drawn correctly
+// We change the Height and Width for 90 and 270 roations so the selection, resize and bisection code still works
+
 import { InitiativeData } from "@/game/comm/types/general";
 import { ServerAsset } from "@/game/comm/types/shapes";
 import { GlobalPoint } from "@/game/geom";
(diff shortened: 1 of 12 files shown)


client/src/game/shapes/asset.ts
```diff
@@ -1,3 +1,8 @@
+// Developer Note: Rotation
+// We only allow 0, 90 ,180 and 270 degrees
+// On an Asset, we need to keep the rotationDegrees to ensure the image is drawn correctly
+// We change the Height and Width for 90 and 270 roations so the selection, resize and bisection code still works
```
Kruptein, line 4:
```suggestion
// We change the Height and Width for 90 and 270 rotations so the selection, resize and bisection code still works
```

client/src/game/shapes/asset.ts
```diff
@@ -27,7 +32,35 @@ export class Asset extends BaseRect {
     draw(ctx: CanvasRenderingContext2D): void {
         super.draw(ctx);
         try {
-            ctx.drawImage(this.img, g2lx(this.refPoint.x), g2ly(this.refPoint.y), g2lz(this.w), g2lz(this.h));
+
+            if (this.rotationDegrees == 0) {
+                ctx.drawImage(this.img, g2lx(this.refPoint.x), g2ly(this.refPoint.y), g2lz(this.w), g2lz(this.h));
+            }
+            else {
```
Kruptein, line 39:
```suggestion
            } else {
```

client/src/game/shapes/shape.ts
```diff
@@ -339,6 +360,16 @@ export abstract class Shape {
         if (sync) socket.emit("Shape.Layer.Change", { uuid: this.uuid, layer, floor: newLayer.floor });
     }
 
+    rotate(degrees: number, sync: boolean): void {
+        const oldLayer = layerManager.getLayer(this.floor, this.layer);
+        this.rotationDegrees = degrees;
+        if (this.rotationDegrees >= 360) this.rotationDegrees = 0;
```
Kruptein, line 366:
```suggestion
        this.rotationDegrees = degrees % 360;
```

client/src/game/shapes/shape.ts
```diff
@@ -339,6 +360,16 @@ export abstract class Shape {
         if (sync) socket.emit("Shape.Layer.Change", { uuid: this.uuid, layer, floor: newLayer.floor });
     }
 
+    rotate(degrees: number, sync: boolean): void {
+        const oldLayer = layerManager.getLayer(this.floor, this.layer);
```
Kruptein, line 364: Why is it called 'oldLayer' ? the layer doesn't change so the naming might seem confusing.

client/src/game/shapes/shape.ts
```diff
@@ -339,6 +360,16 @@ export abstract class Shape {
...
+        this.rotationDegrees = degrees;
+        if (this.rotationDegrees >= 360) this.rotationDegrees = 0;
+        if (oldLayer === undefined) return;
+        oldLayer.invalidate(true);
```
Kruptein, line 368: You can just call `this.invalidate` instead of having to fetch the layer yourself.

server/README.md
```diff
@@ -18,8 +18,6 @@ This requires the client to be ran in development mode as well.
 python planarserver.py dev
 ```
 
-If both server and client are running you can launch PA by visiting http://localhost:8000
-
```
Kruptein, line 22: This should not be removed

server/save.py
```diff
@@ -330,6 +330,15 @@ def upgrade(version):
             )
         db.foreign_keys = True
         Constants.get().update(save_version=Constants.save_version + 1).execute()
+    elif version == 23:
```
Kruptein, line 333: You also need to increase the SAVE_VERSION at the top of this file, or the migration will never happen

client/src/game/shapes/asset.ts
```diff
@@ -27,7 +32,35 @@ export class Asset extends BaseRect {
...
+                    ctx.translate(-(g2lx(this.refPoint.x) + (g2lz(this.w) / 2)), -(g2ly(this.refPoint.y) + (g2lz(this.h) / 2)))
+
+                // draw the image
+                if (this.rotationDegrees == 90 || this.rotationDegrees == 270)
+                    ctx.drawImage(this.img, g2ly(this.refPoint.y), g2lx(this.refPoint.x), g2lz(this.h), g2lz(this.w));
+                else
+                    ctx.drawImage(this.img, g2lx(this.refPoint.x), g2ly(this.refPoint.y), g2lz(this.w), g2lz(this.h));
+
+                // we�re done with the rotating so restore the unrotated context
+                ctx.restore();
```
Kruptein, line 62: Although I'm sure this works, I think it's faster to use something like this:
```javascript
ctx.setTransform(1, 0, 0, 1, x, y);
ctx.rotate(rotation); // in radians
ctx.drawImage(image, -image.width / 2, -image.height / 2);
ctx.setTransform(1,0,0,1,0,0); // reset the transform
```
As it does not require context saving/restoring.  What do you think ?

client/src/game/shapes/shape.ts
Milamberz: Copied and pasted from the Shape.moveFloor() and didn't rename it, will fix.

client/src/game/shapes/shape.ts
Milamberz: done

server/save.py
Milamberz: done

client/src/game/shapes/shape.ts
Milamberz: No longer used, swapped to this.invalidate(true)

server/README.md
Milamberz: Reverted (I hope).

Milamberz: Reverted (I hope, still learning git)

client/src/game/shapes/asset.ts
Milamberz: I played with a few methods, and started down a few paths until I hit the selection code, which lead to this solution. Mainly so I didn't have to overhaul selection and resizing to deal with a rotated Asset.

I haven't done much with canvas so this might not be the best solution (but it works).

Rotating from the center point doesn't quite work. Snap to grid works fine with even number of tiles, but on odd number of tiles it leaves the image offset by half a tile.

Not sure I see what your doing until I run it in code. Will review.

server/save.py
Milamberz: SAVE_VERSION needs to be +1 from the last elif? 
i.e. SAVE_VERSION = 24
Done.....

Kruptein: Yes on startup the server checks if a migration is required based on the SAVE_VERSION :)

Kruptein:
if you can't make the code I gave earlier working that's fine.  I would like to offer some more finegrained rotations in the future, but as I have nothing at all now, this is much appreciated.

There are a bunch of linting warnings, (I hoped they would be flagged as errors need to check my CI config).  So please fix them as well (most should be auto fixable with `npm run lint` in the client folder.

Milamberz:
Now that I'm looking at it better, I think I should do it again. 
Incremental rotation needs the maths behind it to work properly, and it has big impacts for rotation of grouped shapes, selection etc.

I'll work on design/maths first and run it by you.

PR #300, opened 2020-04-13
feature(Token): Cycle owned tokens with spacebar

Kruptein:
By pressing spacebar you cycle between your owned tokens. (i.e. center the screen on the token)


10leej:
Just curious does this limit selection to tokens on the same layer?

Kruptein:
It goes over all your owned tokens irregardless of layer or floor currently

PR #310, opened 2020-04-19
Add example configuration

jirsat:
Two new example files.
Nginx configuration and systemd service file.
Especially nginx configuration is tricky due to socket.io so including it in repository should help point people to the right direction.
This should be an addition to documentation, not a replacement.

### DIFF
--- a/examples/PlanarAlly.nginx
+++ b/examples/PlanarAlly.nginx
@@ -1,0 +1,75 @@
+upstream pa {
+   server 0.0.0.0:8000 fail_timeout=0;
+
+## If you use unix socket comment the previous and uncomment the next line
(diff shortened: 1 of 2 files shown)


Kruptein:
Hey, I don't know whether you noticed, but I declined your earlier PR, because you should make documentation changes on my other repository: https://github.com/Kruptein/planarally-docs.

As that's the location that populates the https://planarally.io website.

jirsat:
I've noticed, but this is something I think should be included. I'm working on documentation too but this will help a lot i think. But if you thing this should be done other way please tell, I'll keep it in mind.

Kruptein:
Can you add a changelog file ?

jirsat:
Do you still need this or should i close it? Sorry for inactivity

PR #323, opened 2020-04-27
Delete server files removed with asset manager

chisaipete:
Relatively simple two line change to enable the server to actually delete files. Folders and Files that are removed by the Asset Manager call the same `Asset.Remove` endpoint, so it is necessary to check if the asset being removed is a file or folder. Folders aren't created on the server, so checking for the file_hash's existence is sufficient to protect against errors which arise from trying to delete non-existent assets on the server.

I've briefly tested this and confirmed that I can add and remove assets and their asset files indeed are removed from the server disk.

### DIFF
--- a/server/api/socket/asset_manager.py
+++ b/server/api/socket/asset_manager.py
@@ -117,6 +117,11 @@ async def assetmgmt_rm(sid, data):
         return
     asset.delete_instance(recursive=True, delete_nullable=True)
 
+    if asset.file_hash is not None and (ASSETS_DIR / asset.file_hash).exists():
+        if Asset.select().where(Asset.file_hash == asset.file_hash).count() == 0:
+            logger.info(f"No asset maps to file {asset.file_hash}, removing from server")
+            (ASSETS_DIR / asset.file_hash).unlink()
+
 
 @sio.on("Asset.Upload", namespace="/pa_assetmgmt")
 @auth.login_required(app, sio)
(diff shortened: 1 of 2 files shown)


Kruptein:
Pretty straight forward, just one remark;

It's possible that some people have the exact same file uploaded, which will result in the same file hash.  you want to make sure that there is no other Asset that maps to the file before removing it

server/api/socket/asset_manager.py
```diff
@@ -115,6 +115,10 @@ async def assetmgmt_rm(sid, data):
     if asset.owner != user:
         logger.warning(f"{user.name} attempted to remove a file it doesn't own.")
         return
+    if asset.file_hash is not None and (ASSETS_DIR / asset.file_hash).exists():
+        unique_asset_owners = Asset.select(Asset.owner).where(Asset.file_hash == asset.file_hash).distinct()
```
Kruptein, line 119: You can use `.count()` I believe at the end to do the count in database, so no need to do the length check below.

I can see why you check distinct on the asset owner, but I think you just want to make sure there isn't any other asset that links to it irregardless of the owner.  Suppose you've uploaded a map to two folders and decide you don't need one of them, deleting the one in folder x/y/z should not affect the one in a/b/c.

chisaipete: Yeah, that makes sense, I was trying to be too specific and it cluttered the code and didn't really work 😄 I'll take another shot at it

chisaipete:
Okay, this solution feels much cleaner and seems to work without issue.

Issue #92, opened 2018-11-06
Add translation options


Kruptein:
Look into providing language selection for the UI.

Possibly using i18next

Kruptein:
https://www.codeandweb.com/babeledit/tutorials/how-to-translate-your-vue-app-with-vue-i18n seems like a good resource to integrate i18n in PA.

It has an attractive set of features (pluralization, grouping, extending...)

dthv:
Just scrolled by, accidentally: This can be closed? :smile:

PR #397, opened 2020-06-05
Get i18n ready and provide language selection

hmqgg:
#92 suggests using [vue-i18n](https://github.com/kazupon/vue-i18n) to implement i18n.

To improve the localization experience, I added [vue-i18n-extract](https://github.com/pixari/vue-i18n-extract/) dev package to extract i18n ready strings from `.vue` files.
To extract strings and add missing keys to translation files, exec this under `client` folder:
~~`
npx vue-i18n-extract report --vueFiles './src/**/*.vue' -l './src/locales/*.json' -a
`~~
Update: use i18n:report
```bash
npm run i18n:report
```
and add missing keys manually.

Todo: 
- [x] Replace existing strings with `$t("to be translated")`
- [x] ~~Language selection UI~~
- [x] ~~Language selection stored (maybe using web localStorage, or DB for syncing options)~~
- [x] ~~`zh` translation~~
- [x] ~~PR reminder: use i18n instead of plain text and execute string extract before merge~~
- [x] ~~When there are too many translation files, implement lazy loading~~

Separate PR for UI things and language translations, this PR is only for i18n.


hmqgg:
Maybe we could use structure to make the translation file clear, e.g.
In the translation JSON file:
```JSON
{
  "login": {
    "server_version": "Server version:"
  },
  "errors": [
    "something happened",
    "login failed"
  ]
}
```
and to use:
```
$t("login.server_version")
$t("errors[0]")
```

hmqgg:
Another issue is the back-end message translation, e.g. if `login.vue` requested login API but something happened, the API returned an error message from plain text in the back-end codes.

hmqgg:
Using a lot of switch-case statement to translate, because I want to make it extractable.

If passed it directly to i18n, the extractor couldn't get it and we'll need more works to record those translation keys and adding them to the translation files.

hmqgg:
@Kruptein ready to be merged ;)

Kruptein:
Is it much work to use the structure you mentioned earlier where you use things like `login.server_version` ?

Because I do prefer that approach indeed, it just brings some more natural structure to the json file instead of being one large dump.

Otherwise at first glance looks good, giong to look at it more in-depth tomorrow and give it a test run myself

hmqgg:
> Is it much work to use the structure you mentioned earlier where you use things like `login.server_version` ?
> 
> Because I do prefer that approach indeed, it just brings some more natural structure to the json file instead of being one large dump.
> 
> Otherwise at first glance looks good, giong to look at it more in-depth tomorrow and give it a test run myself

Not much, I'd like to organize them tomorrow

Pros for non-structural:
- For most of English text, we could just copy/paste(regex replace all at once) to generate the translation file from raw keys.
- Hinting linguister what the translation key is, without looking up code or `en.json`.

Cons:
- Hard to organize and maintenance, especially when considering more and more features. And adding new strings to existing `.vue` file will make keys out of order, kind of annoying.
- Since the key having their own meaning and it might be a long sentence, once it needs change the wording or something like that, all the translation files need to change the key.

client/output.json
```diff
@@ -0,0 +1 @@
..."line":89,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.username","line":91,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.email","line":104,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.danger_zone","line":116,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.new_pwd","line":118,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.repeat_pwd","line":124,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.common.cancel","line":134,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.change_pwd","line":141,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.delete_account","line":147,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.account.delete_confirm_msg","line":150,"file":"/src/settings/account.vue","language":"en"},{"path":"settings.settings.account_settings","line":50,"file":"/src/settings/settings.vue","language":"en"},{"path":"settings.settings.account","line":60,"file":"/src/settings/settings.vue","language":"en"}],"unusedKeys":[]}
```
Kruptein, line 1: All good, just these keys are missing a translation key in the english locales if I'm not mistaken

hmqgg: Deleted, sorry for bringing unnecessary files.

Kruptein: But they are missing in en.json right ?

hmqgg: Nope, it's just generated by `i18n:report`, and I just resolved all those missing keys.

PR #401, opened 2020-06-06
feat(I18n): Add localization for Chinese

hmqgg:
Add Chinese localization, it could be used for testing language switch.

Update the changelog after implement language selection GUI?


Kruptein:
Feel free to already add it to the changelog, if people ask about it before they can do it in the UI we can guide them.

PR #418, opened 2020-06-08
[FEAT] I18n language selection UI and store locale

hmqgg:
`client/src/core/components/languageDropdown.vue` implements a dropdown selection for language switching.
It hard-coded supported languages and I do think it's acceptable for the current use case since the translation file is webpacked.
Maybe it needs a better UI/UX, but I'm not good at this.

After language selected, it will call localStorage API and stores the locale info for next time loading such as refreshing.

### DIFF
--- a/client/src/auth/login.vue
+++ b/client/src/auth/login.vue
@@ -2,10 +2,16 @@
 import Vue from "vue";
 import Component from "vue-class-component";
 
+import LanguageDropdown from "@/core/components/languageDropdown.vue";
+
 import { coreStore } from "@/core/store";
 import { postFetch } from "../core/utils";
 
-@Component
+@Component({
+    components: {
+        languageDropdown: LanguageDropdown,
+    },
+})
 export default class Login extends Vue {
     username = "";
     password = "";
@@ -98,6 +104,11 @@ export default class Login extends Vue {
                     <span>
                         <i aria-hidden="true" class="fas fa-lock"></i>
                     </span>
+                </div>
+
+                <div class="input">
+                    <label for="langDropdown" v-t="'locale.select'"></label>
+                    <languageDropdown id="langDropdown" />
                 </div>
 
                 <div style="display:flex;">
--- a/client/src/core/components/languageDropdown.vue
+++ b/client/src/core/components/languageDropdown.vue
@@ -1,0 +1,34 @@
+<script lang="ts">
+import Vue from "vue";
+import Component from "vue-class-component";
+
+import { Prop } from "vue-property-decorator";
+
+@Component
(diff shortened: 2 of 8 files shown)


client/src/auth/login.vue
```diff
@@ -100,6 +106,11 @@ export default class Login extends Vue {
                     </span>
                 </div>
 
+                <div class="input">
+                    <label for="langDropdown" v-t="'locale.select'"></label>
+                    <languageDropdown id="langDropdown" />
```
Kruptein, line 111: There is no need to specify an `id` if you never use it in css/js

client/src/core/components/languageDropdown.vue
```diff
@@ -0,0 +1,34 @@
+<script lang="ts">
+import Vue from "vue";
+import Component from "vue-class-component";
+
+import { Prop } from "vue-property-decorator";
+
+@Component
+export default class LanguageDropdown extends Vue {
+    @Prop() value!: string;
```
Kruptein, line 9: This seems to be unused?

client/src/i18n.ts
```diff
@@ -17,7 +17,7 @@ function loadLocaleMessages(): LocaleMessages {
 }
 
 export default new VueI18n({
-    locale: process.env.VUE_APP_I18N_LOCALE || "en",
+    locale: localStorage.getItem("locale") || process.env.VUE_APP_I18N_FALLBACK_LOCALE || "en",
```
Kruptein, line 20: Why did this become fallback locale ?

```suggestion
    locale: localStorage.getItem("locale") || process.env.VUE_APP_I18N_LOCALE || "en",
```

client/src/locales/de.json
```diff
@@ -1,4 +1,10 @@
 {
+    "locale": {
+        "select": "Select Language:",
+        "en": "English",
+        "zh": "中文",
+        "de": "Deustch"
```
Kruptein, line 6:
```suggestion
        "de": "Deutsch"
```

client/src/locales/en.json
```diff
@@ -1,4 +1,10 @@
 {
+    "locale": {
+        "select": "Select Language:",
+        "en": "English",
+        "zh": "中文",
+        "de": "Deustch"
```
Kruptein, line 6:
```suggestion
        "de": "Deutsch"
```

client/src/locales/de.json
hmqgg: typo

client/src/core/components/languageDropdown.vue
hmqgg: Removed

Issue #427, opened 2020-06-13
[BUG] Colour of aura disappears when token is out of frame


Mikkelhauvgaard:
**Describe the bug**
When a token with an aura, that has a non-zero alpha value, is out of the visible frame, the colour of the aura will disappear, resulting in the light showing having no colour.

**To Reproduce**
1. Create a new token
2. Create aura on said token and change it's alpha to a non-zero value
3. Pan the frame such that the token is not visible but the aura should be
4. The aura now loses it's color

**Expected behavior**
The aura should reatin it's color even when out of frame

**Screenshots**
The aura has colour when in frame
![image](https://user-images.githubusercontent.com/19272763/84569233-42ba9080-ad85-11ea-9321-7ea415bc6c9d.png)
The colour disappearing when out of frame
![image](https://user-images.githubusercontent.com/19272763/84569237-4ea65280-ad85-11ea-9fc9-04dbea8fe609.png)

**Desktop:**
 - OS: Windows/Linux
 - Browser: Firefox
 - Firefox version: 77.0.1
 - PlanarAlly version: 0.21.0

PR #429, opened 2020-06-13
Fixed visibleInCanvas to support aura

Mikkelhauvgaard:
- [x] Make sure to do your PR on the `dev` branch and NOT on the `master` branch.  This repo uses gitflow which only uses the master branch as a release branch.

- [x] Make sure to update the `CHANGELOG`.
Implemented visibleInCanvas for CircularTokens, thus fixing the bug where auras would not be displayed when the source token is out of frame #427

### DIFF
--- a/client/src/game/shapes/shape.ts
+++ b/client/src/game/shapes/shape.ts
@@ -94,13 +94,13 @@ export abstract class Shape {
 
     abstract center(): GlobalPoint;
     abstract center(centerPoint: GlobalPoint): void;
-    visibleInCanvas(_canvas: HTMLCanvasElement): boolean {
+    visibleInCanvas(canvas: HTMLCanvasElement): boolean {
         for (const aura of this.auras) {
             if (aura.value > 0 || aura.dim > 0) {
                 const r = getUnitDistance(aura.value + aura.dim);
                 const center = this.center();
                 const auraArea = new BoundingRect(new GlobalPoint(center.x - r, center.y - r), r * 2, r * 2);
-                if (auraArea.visibleInCanvas(_canvas)) return true;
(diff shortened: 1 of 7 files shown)


Kruptein:
Looks fine, but this only fixes it for circular tokens and not for other shapes.

Does reinstantiating the bas code for visibleInCanvas in shape fix it ? I need to check why I commented it in the first place

client/src/game/shapes/shape.ts
```diff
@@ -95,12 +95,14 @@ export abstract class Shape {
...
-        //         const auraCircle = new Circle(this.center(), aura.value);
-        //         if (auraCircle.visibleInCanvas(canvas)) return true;
-        //     }
-        // }
+        for (const aura of this.auras) {
+            if (aura.value > 0 || aura.dim > 0) {
+                const r = getUnitDistance(aura.value + aura.dim);
+                const center = this.center();
+                const auraArea = new BoundingRect(new GlobalPoint(center.x - r, center.y - r), r * 2, r * 2);
+                if (auraArea.visibleInCanvas(_canvas)) return true;
```
Kruptein, line 103:
```suggestion
                if (auraArea.visibleInCanvas(canvas)) return true;
```

client/src/game/shapes/shape.ts
```diff
@@ -95,12 +95,14 @@ export abstract class Shape {
     abstract center(): GlobalPoint;
     abstract center(centerPoint: GlobalPoint): void;
     visibleInCanvas(_canvas: HTMLCanvasElement): boolean {
```
Kruptein, line 97:
```suggestion
    visibleInCanvas(canvas: HTMLCanvasElement): boolean {
```

Issue #435, opened 2020-06-18
[BUG] dialog background is cut at the right side in Firefox and some other browsers


Goury:
**Describe the bug**
Subj

**To Reproduce**
1. Open a token's properties in Firefox.
2. Look at it
3. Background is missing on the right side

**Expected behavior**
To have background for the entire dialog

**Screenshots**
![Снимок экрана от 2020-06-18 12-43-45](https://user-images.githubusercontent.com/1081221/85005432-7456a180-b161-11ea-9b63-f7656e3a94f3.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04 on client, Ubuntu 16.04 on server
 - Firefox 68-77
 - 0.21.0

**Additional context**
*additional context was removed by Kruptein's request*

dthv:
I can confirm this. This did not happen until 0.20.0 (i.e. not in 0.19.0) and, thus, could be related to the overhaul of shape ownership.

Goury:
I remember this bug since 0.17.
Just found few minutes to fix it finally.

Kruptein:
I don't have this on my ff either (ff 77).

Kruptein:
This seems to be an issue specific to firefox on ubuntu.  Both the firefox on my windows and the firefox on my archlinux machines render completely correct.  I also have confirmation that it renders perfeclty fine on macOS on firefox.

Is there anything you have in your settings that affects styling, different fonts or the like.  there is some reason why your ff or ff on ubuntu derivatives render differently and I'm less inclined to just throw over the styling of all other places before first figuring out if some other issue is at hand here.  Can you figure out what special styling ubuntu does that might be the source of rendering this differently.

Also I don't really appreciate the tone of the additional context statement. It's not like I'm doing this on purpose.  I test myself regularly on firefox, I have friends that use firefox as their main driver that both DM and play.

Goury:
GTK theme for inputs, most likely.
But why limit max-width anyway?
I don't see anything bad in being able to resize this dialog.

Kruptein:
Because long usernames then impact the width of the entire dialog.

This is an ubuntu/gtk problem.  I'll find out myself what kind of extra styling they do that shouldn't happen.

Goury:
Should limit .owner and select max width and also .owner max height and also make it wrap words in this case.

Limiting entire dialog width just because some elements may be too long is wrong and leads to problems like this.  
[...]

PR #436, opened 2020-06-18
#435 fix (dialog background is cut at the right side in Firefox and some other browsers)

Goury:
Got your [#435](https://github.com/Kruptein/PlanarAlly/issues/435) quickfixed


Kruptein:
The UI rework of the edit dialog should have fixed this problem in another way.

Issue #437, opened 2020-06-18
[BUG] map is being scrollable when left menu is open in FIrefox


Goury:
**Describe the bug**
Subj

**To Reproduce**
1. Click on that cog icon in upper left corner
2. A scroll bar or two (in smaller window) appear
3. Using scroll wheel destroys experience now
4. Tremble in fear

**Expected behavior**
No scroll bars nor scrolling should happen

**Screenshots**
![Снимок экрана от 2020-06-18 13-13-07](https://user-images.githubusercontent.com/1081221/85008670-f34dd900-b165-11ea-8da9-7f17f7f0a9e4.png)
![Снимок экрана от 2020-06-18 13-14-02](https://user-images.githubusercontent.com/1081221/85008691-f9dc5080-b165-11ea-94f0-026e7cb94fd0.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04 on client, Ubuntu 16.04 on server
 - Browser Firefox 68-77
 - Version 0.21.0

**Additional context**
Can we please not Ignore Firefox in the future?

Kruptein:
I'll take a look at the change, but my firefox does not have this issue, under which circumstances does this trigger? are there some specific requirements? I'm on ff 77

Kruptein:
Ok so it only triggers for campaigns with more locations than the horizontal screen can fit.  I fixed it earlier but hadn't properly tested it with the left sidebar open.

The fix in your PR just drops a bunch of the locations by hiding them which is not what we want.

The proper fix is for the #locations width calculation in location.vue needs to take the sidemenu into account for its calculation.

Goury:
@Kruptein this is why I called it "quickfix".  
It doesn't fix logic, just appearance.  
Sure it's better to have it actually properly fixed.  
But I have a game session tomorrow morning and some one else might have one too, so they can quickgrab this quickfix and have at it.

Kruptein:
I mean, you can do whatever you want on your own server, if you want to quickfix it there do as you please, but I'm not going to 'quickfix' this when it's incorrect and not critical.

Goury:
I don't insist, I just shared it because I felt like sharing it

PR #438, opened 2020-06-18
Upstream ff overflow fix

Goury:
Got your #437 quickfixed


Kruptein:
This fix is incorrect as it hides locations, see the issue discussion for more info.

Also please use more descriptive titles, I've had to go back and forth between your two PRs because they are pretty non descriptive (mostly the other one).

Issue #434, opened 2020-06-17
[BUG] CentOS 8 install fails.


theone85ca:
I apologize, I'd check Discord or the link for support but I wasn't sure where that might be.

**Describe the bug**
Installing using the instructions at the link below seems to be successful, however after running the planarsever.py and attempting to load the site, the following error is generated and the site never loads.

I'm also not able to find the folder or files associated with this error. There is no /templates folder inside of the server folder.

`FileNotFoundError: [Errno 2] No such file or directory: 'templates/index.html'`

Instructions: https://www.planarally.io/tutorial/setup/self-hosting

**To Reproduce**
1. git clone the repo into a folder on a fresh CentOS8 install.
2. cd in to the server directory and run `pip3 install --user -r requirements.txt`
3. run `python3 planarserver.py`
4. Load the page from the servers public IP address
5. Observe the error on the server and receive an `ERR_EMPTY_RESPONSE` message in Chrome.

**Expected behavior**
PlanarAlly should load in the browser with no errors in the console.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: CentOS 8
 - Browser: Chrome
 - Python Version: 3.6.8

dthv:
cf #218 – if you clone the repository, you first have to build the js files by yourself.

You can also use a binary file of the latest release.

Kruptein:
Exactly what dthv said.  If you clone the repo you'll also need to build the client.  If you don't want to do that use one of the `-bin` release files which have the client prebuild

Issue #218, opened 2020-02-25
Server error: Unable to find 'templates\\index.html'


AJAnderson:
Hello,

I did the following:
1. clone the dev branch
2. ran "pip install -r requirements.txt", 
3. fired up the server via python planarserver.py (both dev and normal)
4. tried to connect using chrome and 'http://localhost:8000/'

I get the error message on the server cmd line window "FileNotFoundError: [WinError 3] The system cannot find the path specified: 'templates\\index.html'

I tried to switch to the master repo and copy down the index.html I found in that repo but all I get is a blank screen.

Any suggestions?

Kruptein:
Yes, I changed some things around and now you have to build the js files themselves if you don't run the binary.

I can foresee some people with enough python knowledge but no js knowledge running into some issues with this, so what I suggest is that I'll always attach a zip with the js artifacts to the releases.

For the non-master git checkouts though, you'll need to take care of this yourself. Check the [CONTRIBUTING](https://github.com/Kruptein/PlanarAlly/blob/dev/CONTRIBUTING.md) document.  I'll make sure to create some more clear installation documents as well :)

PR #440, opened 2020-06-19
Gh 434 warning

dthv:
To (further) prevent bug-reports like #434 or #218 the server now provides a helpful error-message, pointing towards the tutorial.

Tested on Linux; not sure whether concatenation in line 10 actually should be done with '/' instead of '+' (does os.isdir() support that?) to support Windows denoting directories with backslash, though.

### DIFF
--- a/server/planarserver.py
+++ b/server/planarserver.py
@@ -7,7 +7,7 @@ This is the code responsible for starting the backend and reacting to socket IO 
 import sys
 from utils import FILE_DIR
 
-if( (FILE_DIR / "templates").exists() is False):
+if not (FILE_DIR / "templates").exists():
     print('You must gather your par— you must build the client, before starting the server.\nSee https://www.planarally.io/tutorial/setup/self-hosting/ on how to build the client or import a pre-built client.')
     sys.exit(1)
 
--- a/server/planarserver.py
+++ b/server/planarserver.py
@@ -3,6 +3,14 @@ PlanarAlly backend server code.
 This is the code responsible for starting the backend and reacting to socket IO events.
 """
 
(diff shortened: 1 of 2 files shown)


server/planarserver.py
```diff
@@ -3,6 +3,14 @@
 This is the code responsible for starting the backend and reacting to socket IO events.
 """
 
+# Check for existence of './templates/' as it is not present if client was not built before
+import os
+import sys
+
+if(os.path.isdir( os.path.dirname( os.path.realpath(__file__) ) + '/templates') is False):
```
Kruptein, line 10:
```suggestion
from utils import FILE_DIR

if (FILE_DIR / "templates").exists():
```
Could you try it with the above change instead?
Although not wrong, I prefer the newer pathlib API over the original os.path API.
Also the FILE_DIR from utils will work with windows executables as well, as the path is different in that case.

dthv:
Again, only tested on Linux.

server/planarserver.py
```diff
@@ -3,6 +3,14 @@
 This is the code responsible for starting the backend and reacting to socket IO events.
 """
 
+# Check for existence of './templates/' as it is not present if client was not built before
+import sys
+from utils import FILE_DIR
+
+if( (FILE_DIR / "templates").exists() is False):
```
Kruptein, line 10:
```suggestion
if not (FILE_DIR / "templates").exists():
```
Another small nitpick, it's more pythonic to use not here instead of an is False.  also there is no need to add surrounding parentheses in python

dthv: Thanks! :D

PR #446, opened 2020-06-27
Added mysql support

Hauvgaard:
I've added support for mysql as the backend database and made it relatively easy to add support additional databases.


develroo:
Cool. I prefer postgres myself, but this would make exporting data much easier, I think. Assets and data can be just bundle up with an SQL schema and voila, we could import and export user data and assets.

Kudos!

PR #472, opened 2020-07-21
Lock/Unlock button on selectInfo

hmqgg:
Show Lock/Unlock on the selectInfo control, and click to switch (requires Edit Access).

PR #474, opened 2020-07-21
update German translation

dthv:
update german translation on #471 and #472


client/src/locales/de.json
```diff
@@ -182,7 +182,10 @@
                 "select_info": {
                     "open_shape_props": "Eigenschaften der Form öffnen",
                     "quick_edit_tracker": "Tracker schnell ändern",
-                    "quick_edit_aura": "Aura schnell ändern"
+                    "quick_edit_aura": "Aura schnell ändern",
+		    "lock": "Feststellen ein/ausschalten",
+		    "new_value_NAME": "Neuer Wert von {name}:",
+		    "updating_NAME": "{name} aktualisieren"
```
Kruptein, line 188: The indentation is wrong here

dthv: I have no frickin' idea why vim keeps doing this. we had this also last time. /o\

Issue #311, opened 2020-04-21
Map resize does not replicate


dthv:
I'm using Planar-Ally 0.19.3 on a Linux basis (please ask specifics if you need them), accessing with Firefox 75.0. My players use different browsers, I think, but I know of Firefox being used, as well. I'll ask them for specifics, if you need them.

Map resize functions quite well, but would not replicate to my players. Thus, while my map is aligned with the grid, the one they see, is not.

10leej:
I don't have this problem and am unable to replicate the issue. I would naturally assume this could be an issue on the browser side.  
Try refreshing the page (players included), clearing browser cache, including rebooting the server instance/docker container.

dthv:
Refreshing pages was tried multiple times by everyone, did not fix.
If it's an issue on the browser side, most likely on GM's side, because no one out of five players could see the adjusted map.

Does PA need any cookies etc. that could be prevented/hindered by strict privacy settings?

I will try clearing browser cache and server reboot asap and report back. Thanks for your fast reply!

10leej:
PA does not use or save cookies as far as I know. I do know that asset placements are saved in the /data/planar.sqlite file could be an error occurred there but outside of nuking and remaking it from scratch I'm not certain on a fix for that.

Kruptein:
Is there anything in the web console of the DM (or the players) or in the server log that is popping up ?

dthv:
I've just checked: Nope.
I'm not sure what you mean by "web console", though. Server log is empty (on this specific issue).

Kruptein:
Somehow forgot about this issue sorry for that!, Has this problem persisted in other sessions?

dthv:
Sort of. Resize actually does replicate if it's done while the players are not logged-in in the session. So, as work-around, I either prepare maps in advance (which is always better using a VTT, anyway) or ask the players to log-off for me to upload and resize a map. When resize is done _before_ the player downloads the map, it seems to work fine.

This sounds even more like an issue with browser cache, but when I tried, re-loading while clearing cache did not work.

> Has this problem persisted in other sessions?
Well, I did not open a whole new session to try out whether it persists only to a certain session. I will take some time in a few days and give it thorough testing. I will probably upgrade to 0.20.0, though, as the release-notes alone are GREAT. :)

dthv:
[...]

PR #475, opened 2020-07-21
bug(Map): Fix map resize not replicate

hmqgg:
#311 reproducible after PR #451 event-improvements.

Make map tool call `Shape.Update` in `apply()` might fix this.

*Needs further investigation*


Kruptein:
Seems good, I would just like a changelog entry :)

hmqgg:
> Seems good, I would just like a changelog entry :)

Ready to merge 😊

PR #478, opened 2020-07-22
[FEAT] Dragging asset while pressing shift key scales it to one grid

hmqgg:
Makes it easier to drag and drop tokens. And it will keep the aspect ratio.


Kruptein:
The only thing I would suggest is to use ctrlKey instead of shiftKey as ctrl is already used to retain aspectratio on resize of shapes

PR #480, opened 2020-07-22
feat(Ruler): Improve ruler distance text readability

hmqgg:


client/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "core-js": "^3.6.5",
+    "invert-color": "^2.0.0",
```
Kruptein, line 17: There is no need for a specialized invert color library.  tinycolor2 is already a dependency which is very flexible.  It can give the most readable color for example.

client/src/game/shapes/text.ts
```diff
@@ -3,22 +3,26 @@ import { BoundingRect } from "@/game/shapes/boundingrect";
 import { Shape } from "@/game/shapes/shape";
 import { ServerText } from "../comm/types/shapes";
 import { rotateAroundPoint } from "../utils";
+import invert from "invert-color";
 
 export class Text extends Shape {
     type = "text";
     text: string;
     font: string;
+    readable = false;
```
Kruptein, line 12: I'm a bit conflicted on this.  The first issue is that this is not synced to the server, which is fine for a ruler on your own client, but will show up differently on another client for example.

Additionally this conflicts with text objects that actually do set a strokeColour.
What's wrong with setting the stroke colour on the text object for the ruler ? If you really want the shadow effect you can set it to the same colour as the stroke colour if it is set.

What do you think ?

hmqgg: Yep, text with strokeColour will conflict with this, I didn't realize there is before.

Kruptein:
Sorry for the later review but the resize problem was keeping me busy.  I tested it out looks clear! just remove the package-lock you still have lingering ;)

PR #569, opened 2020-11-18
use numpad for viewport navigation and token movement; fix ctrl to ce…

dthv:
…nter viewport on origin

### DIFF
--- a/client/src/game/input/keyboard.ts
+++ b/client/src/game/input/keyboard.ts
@@ -1,4 +1,4 @@
-import { Vector } from "@/game/geom";
+import { GlobalPoint, Vector } from "@/game/geom";
 import { layerManager } from "@/game/layers/manager";
 import { copyShapes, deleteShapes, pasteShapes } from "@/game/shapes/utils";
 import { DEFAULT_GRID_SIZE, gameStore } from "@/game/store";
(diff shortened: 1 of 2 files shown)


client/src/game/input/keyboard.ts
```diff
@@ -116,9 +160,24 @@ export function onKeyDown(event: KeyboardEvent): void {
...
             gameStore.toggleUI();
         } else if (event.key === "0" && event.ctrlKey) {
-            // Ctrl-0 - Re-center/reset the viewport
-            gameStore.setPanX(0);
-            gameStore.setPanY(0);
+            // Ctrl-0 or numpad 0 - Re-center/reset the viewport
+            gameManager.setCenterPosition(new GlobalPoint(0, 0));
+            sendClientLocationOptions();
+            layerManager.invalidateAllFloors();
+        } else if (event.keyCode == 101) {
```
Kruptein, line 167: for individual keys it's better to use `event.key` or `event.code`, `event.keycode` is deprecated but slightly easier to work with when using ranges like 37->40.  You can use something like https://keycode.info to find the correct strings.
In the case of numpad it's probably recommended to use event.code as that differentiates between number row and numpad. (you could also use event.location to differentiate)

client/src/game/input/keyboard.ts
```diff
@@ -116,9 +160,24 @@ export function onKeyDown(event: KeyboardEvent): void {
...
+            let targetX = 0;
+            let targetY = 0;
+            if (layerManager.hasSelection()) {
+                const selection = layerManager.getSelection();
+                for (const sel of selection) {
+                    targetX += sel.refPoint.x;
+                    targetY += sel.refPoint.y;
+                }
+                targetX *= 1 / selection.length;
+                targetY *= 1 / selection.length;
```
Kruptein, line 178: just use `/=` instead of `*= 1/`

Kruptein (changes requested):

dthv:
Thanks for the improvements, edited as requested. :slightly_smiling_face:

PR #609, opened 2020-12-18
feat(Map): Add aspect ratio locking and pixel dimension changing

### LINKED ISSUE #424: [FEAT] Make Map Tool able to resize whole asset
Virtually always, one has to resize maps to seamlessly fit under the grid. This has greatly improved by the last rework of the Map Tool, but not yet perfect. Especially on bigger maps, an offset of just one or two pixels can mean increasing inaccuracy of the fitting between map and grid towards the lower right corner of the map.

I know that there are thoughts and talks about a major rework of the asset import system (and am looking forward to that). But until then, I would like to propose a (probably?) much simpler quality-of-life improvement: Make the Map Tool able to resize a whole asset.

This could be a simple checkbox in the "drag an area you wish to resize"-dialogue that, being clicked, immediately leads to the x by y resize dialogue.

I considered the alternative that a dragged area bigger than the actual asset could automatically snap to the size of the asset. But that brings the disadvantage that this will only function with assets that have only "full" squares, i.e. no cut squares on the edges.

### LINKED ISSUE #441: [FEAT] enhance map tool with "keep aspect ratio" functionality
The map tool is designed to help scale a map with a grid to the software's grid. However, some maps don't have grids, but rather a simple legend denoting map scale. This is especially a thing for (large) overland maps. Given that those maps tend not to be square, but often in landscape (and sometimes portrait) format, it's not easy to scale such a map to grid with the tool.

With a 'keep aspect ratio' functionality, this could be greatly improved.

The preferred solution would be to drag the map tool selection area along the scale on either the x or y axis, inputting the appropriate value into the respective input field and then toggling off the other input with a "keep aspect ratio"-button. The software would then scale according to the given value, ignoring the other input and keeping aspect ratio of the asset, instead.

Since there are two buttons needed (one to make the software ignore either x or y) or one button with three possible states (consider x, keep aspect ratio; consider y, keep aspect ratio; consider both x/y), an easy solution would be to interpret a given value of '0' to equal 'ignore this value and keep aspect ratio'. Preferred solution would be, to have both.

Kruptein:
The map tool now has a small aspect ratio lock/unlock widget that can be toggled.

The map tool now also offers you the option to skip the region selection and immediately configure the grid settings and/or pixel dimensions for the entire shape.

This closes #441 and closes #424


Kruptein:
image

PR #663, opened 2021-02-09
feat(Draw): Erase mode

### LINKED ISSUE #619: [FEAT] Map Eraser
When using multiple floors, there are times that I want to remove part of the floor to reveal the floor below. For example, if players are fighting in a tavern and blow a hole in the floor, I would want to erase that portion of the map layer to reveal whatever is below. So far as I can tell, it is currently only possible to remove entire assets or objects, rather than to erase/add transparency to a portion. 

Either an additional build tool or addition to the drawing toolbar adding an eraser tool. From an implementation perspective, this could either mean erasing pixels from the underling object, or adding a mask that adds transparency where the layer/floor has been erased. 

I've tried using the hide/reveal functionality of the draw tool, but it looks like these only interact with each other, and can't be used to reveal through other objects on a floor. 

**Additional context**
None.

Kruptein:
This adds a new mode to the draw tool for DMs.

When operating in erase mode you can draw shapes as usual which will make everything below them on their floor transparent. 

Just as shapes drawn in reveal or hide mode appear automatically on the 'fow' layer, shapes drawn with this tool will appear automatically on the 'map' layer, although nothing prevents you from moving it to another layer.

These shapes are selectable and actually follow the draw stack. So if you move a particular map asset to the top it will appear over the erased area, great for things like a wooden log that lays over a gaping hole for example!

Potential usecases:
- an explosion destroys a part of the floor and you want to show things on the floor below
- you forgot to remove some part of an asset you uploaded and want to patch it in PA itself

Example image:

the black and red shapes are part of the upper floor map layer, whereas the light blue shape is part of the lower floor map.
The polygon is drawn with the erase tool, but the red shape has been moved to the front.
If I were to remove the light blue shape, the polygon would show the white background of the lower floor instead.

image

This closes #619

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ These usually have no immediately visible impact on regular users
 ### Added
 
 -   Client setting to disable zoom behaviour on scroll
+-   Erase option to draw tool
+    -   This makes anything from the current floor below it in the draw stack in its region transparent
 
 ### Changed
 
--- a/client/src/game/layers/layer.ts
+++ b/client/src/game/layers/layer.ts
@@ -269,6 +269,7 @@ export class Layer {
             // To optimize things slightly, we keep track of the shapes that passed the first round
             const visibleShapes: Shape[] = [];
 
+            // Aura draw loop
             for (const shape of this.shapes) {
                 if (shape.options.has("skipDraw") && shape.options.get("skipDraw")) continue;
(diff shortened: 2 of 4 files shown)


ekerstens:
Looks awesome! I love the additional functionality of ordering the eraser shape to keep certain items visible while hiding others.

Issue #641, opened 2021-01-21
[FEAT] Create Draw Shape from Image/Mask, or from Transparent sections of Image on Map Layer


PalasX:
**Is your feature request related to a problem? Please describe.**
The drawing tools to create shapes that block lights and movement, like walls and doors, are great, but can be time consuming and tedious for very large maps. Anything that takes a human a long time to do, and has to be done repeatedly, is prime territory for automation.

**Describe the solution you'd like**
After dragging an asset to the MAP layer, and resizing according to grid size using the MAP TOOL, provide a means to automatically generate a single solid shape from the transparent parts of the image. Alternatively, let the user upload an image containing transparency to be used as a mask to create a shape.

**Describe alternatives you've considered**
Import a standard library for edge detection on a map asset image to try and create a single or series of shapes for walls? It seems more complicated this way.

**Additional context**
Example Map:
![image](https://user-images.githubusercontent.com/37154966/105404733-4dc97f00-5bf8-11eb-9c47-f2aad7fce39e.png)

Example resulting shape (white being transparent), or secondary mask image asset to be used as a "everything that isnt transparent will block light/vision as per user settings":
![image](https://user-images.githubusercontent.com/37154966/105404895-823d3b00-5bf8-11eb-9c5c-901becb5e52d.png)

I realize this could be a huge undertaking, but hoped that since there was so much python to deal with images already, that sorting out which pixels were and werent transparent would at least be near-trivial. I assume translating a 1bit (black/transparent) image into a shape would be the hard part. It could save DMs a significant amount of time and effort when using randomly generation maps, and people that create their own map art assets could export a "pathing" layer of their image maybe?

Thanks again for all the amazing work you do!

Kruptein:
Hey, this is indeed something that would immensely help DMs all around.

The complexity lies in how the lighting/vision actually is calculated. It's based on an algorithm that uses line segments to construct a triangulation that can be quickly queried.

So for all existing shapes when toggling the 'blocks vision' option, it adds all the lines of the shape to the triangulation and the lines are known because they are either simple shapes or hand-drawn polygons.

I'm sure that I should be able to get the lines from a more complex shape such as your mask, so I'm sure it's feasible, it will just as you mentioned maybe take some more time.

Might look into this in the near future! There are some other things that I would like to tackle first, but this one would be nice to have.

Also nice that you provided an example with a mask ;)

PalasX:
[...]

PR #675, opened 2021-02-14
feat(Vision): Use svg mask to draw vision lines

Kruptein:
Setting up all the walls etc can be an arduous task.  To alleviate that work significantly there now is an option to upload an svg to use as a reference for the walls.

This can be accessed from the Extra settings panel when you've selected an asset.

Some considerations:
- The svg is expected to be somewhat simple, as in just a collection of `<path>` elements. Any styling/scripting/other stuff is not supported and the behaviour of such an svg is undefined when used. (the svg is just scraped for its path info and not used otherwise)
- For obvious reasons, how smaller the svg, the better the performance will be in PA.
- Remove everything that should not be a wall (room numbers, doors, stairs, other furnishing, text etc) This is just meant as a wall mask on top of your regular asset.

It's important to remove doors and windows etc, if you want to make them useable/interactive during gameplay. It's not possible to modify the svg/fow path in planarally as you can when you draw it manually! So it is strongly advised to draw the doors/windows manually so that you can toggle their properties and/or open them.

I've personally been using this [online vector tool](https://www.vectorizer.io) to great success. For best results I think you want to set 'roundness' to none, and if there are multiple colours in the image, try to minimize them as this usualy gives the compacter path, but play with this however you want.

Just realize that everything in the svg that is a path will be used for the visibility triangulation.

Thi is the resulting visibility grid in planar ally when using a simple svg based on the pngs in [this issue](https://github.com/Kruptein/PlanarAlly/issues/641):

image

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -13213,6 +13213,11 @@
             "integrity": "sha512-BapA40NHICOS+USX9SN4tyhq+A2RrN/Ws5F0Z5aMHDp98Fl86lX8Oti8B7uN93L4Ifv4fHOEA+pQw87gmMO/lQ==",
             "dev": true
         },
+        "path-data-polyfill": {
+            "version": "1.0.3",
+            "resolved": "https://registry.npmjs.org/path-data-polyfill/-/path-data-polyfill-1.0.3.tgz",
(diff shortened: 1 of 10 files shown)


ekerstens:
This is a fantastic feature. Something worth noting for anyone else looking for an answer is that this feature interacts with the asset "Blocks vision/light" and "Blocks movement" features. It took me a while to realize while trying to utilize it and thinking it wasn't working.  Is it possible to view the SVG path the way you have in the included image?

PR #695, opened 2021-02-25
Asset upload bug

sevzero:
This change causes the websocket used to upload files to wait for an ack before sending the next slice of the file.

I've tested with large sets of tokens known to cause the upload to fail previously as well as large files and with this change they are uploaded with no issues. Uploads are slightly slower now but not significantly so.


Kruptein:
You're going to have to rebase or merge my changes in dev ;)

I've renamed a bunch of vue components which is probably the conflict

sevzero:
Hmm, I think I've messed things up in git. Will close this PR and submit a new one

PR #703, opened 2021-03-01
Added visual trackers

sevzero:
Adds visual indicators to tokens.

There were quite a lot of changes needed to get this working so please let me know if there are any issues or if I've done anything dumb.


client/src/game/shapes/interfaces.ts
```diff
@@ -4,6 +4,9 @@ export interface Tracker {
     name: string;
     value: number;
     maxvalue: number;
+    draw: boolean;
+    primary_color: string;
+    secondary_color: string;
```
Kruptein, line 9: camelCase these!

In the typescript world we do everything with camelCase except those things that arrive from python which are snake_case, which is why the Server* interfaces have snake_case

client/src/game/shapes/shape.ts
```diff
@@ -452,6 +452,31 @@ export abstract class Shape {
             ctx.lineTo(crossBR.x - r, crossTL.y + r);
             ctx.stroke();
         }
+        // Draw tracker bars
+        let bar_offset = 0;
```
Kruptein, line 456: also here use camelCase

client/src/game/ui/selection/edit_dialog/TrackerSettings.vue
```diff
@@ -96,69 +97,98 @@ export default class TrackerSettings extends Vue {
...
+            <div class="summary" @click="updateTracker(tracker.uuid, {})">
+                <label class="name" :for="'check-' + tracker.uuid">{{ tracker.name }}</label>
                 <div
                     v-if="!tracker.temporary"
-                    :key="'remove-' + tracker.uuid"
                     @click="removeTracker(tracker.uuid)"
                     :disabled="!owned"
                     :style="{ opacity: owned ? 1.0 : 0.3, textAlign: 'center' }"
-                    :title="$t('game.ui.selection.edit_dialog.dialog.remove_tracker')"
+                    :title="$t('game.ui.selection.edit_dialog.dialog.delete_tracker')"
```
Kruptein, line 109: Why the change from remove to delete ?

client/src/game/ui/selection/edit_dialog/TrackerSettings.vue
```diff
@@ -96,69 +97,98 @@ export default class TrackerSettings extends Vue {
...
+                        :disabled="!owned"
+                        :title="$t('common.toggle_draw')"
+                    />
+                </div>
+                <div v-if="tracker.draw">Colours</div>
+                <div v-if="tracker.draw" class="colour">
+                    Primary:
+                    <color-picker
+                        :color="tracker.primary_color"
+                        @input="updateTracker(tracker.uuid, { primary_color: $event }, 2)"
```
Kruptein, line 177: Why the `2` here ? I think this should be `false`

Kruptein (changes requested):
At a first glance looks good!
Mostly some fixes regarding casing.

I still need to try it out in action, but doubt there will be many problems.

From the screenshot you showed earlier, we might want to put the colours on two separate lines, because it's kinda cramped with the 'primary' and 'secondary' on one line

sevzero:
Thanks for taking the time to look through everything. All the feedback is completely valid, it looks like a few things crept in from my debugging efforts so thanks for spotting those.

Nothing looks too difficult to fix so I'll try to get it sorted this evening.

client/src/game/models/shapes.ts
```diff
@@ -96,8 +96,8 @@ export interface ServerTracker {
     value: number;
     maxvalue: number;
     draw: boolean;
-    primary_color: string;
-    secondary_color: string;
+    primaryColor: string;
+    secondaryColor: string;
```
Kruptein, line 100: This should actually remain snake_cased :D because this is the representation straight from the server, it's the only exception where the client code uses snake_case.

There is conversion code for auras to convert to and from server representation.

sevzero: whoops. No problem, I'll look up the conversion code from the auras and do something similar with the trackers

sevzero: also better to ask this now, would you prefer to use 'colour' or 'color'? I went with color as it's most likely to be consistent with other libraries and code but I see you've used colour in the auras so can change it to be consistent with that if you prefer.

Kruptein: I don't mind super hard, but I personally enjoy the British variant more, but the American is more common so that's fine as well.

client/src/game/shapes/template.ts
```diff
@@ -24,7 +25,8 @@ export function applyTemplate<T extends ServerShape>(shape: T, template: BaseTem
     if ("options" in template) shape.options = template.options;
 
     for (const trackerTemplate of template.trackers ?? []) {
-        const defaultTracker = createEmptyTracker();
+        const defaultTracker = trackersToServer(shape.uuid, [createEmptyTracker()])[0];
+        console.log(defaultTracker);
```
Kruptein, line 29: ;)

sevzero: Gah! on it :)

Issue #697, opened 2021-02-25
[FEAT] Copy a location to a new session/room


CalebKoch:
**Is your feature request related to a problem? Please describe.**
I am DMing 2 groups through the same Adventure Path, and so I am setting up a lot of maps multiple times. I would like to be able to create the map once in the first session that gets there, and then have an option to copy this into the other groups session, and have the map, walls, lights, and tokens created there. 

**Describe the solution you'd like**
Have a section of the Location Settings Pop-up that lists all of the sessions that you are an owner of, where clicking one of those sessions, will copy the location, floors, tokens, and drawn items from the selected location into a new location on the selected section.

**Describe alternatives you've considered**
I've considered recreating the map manually, which works, but takes a significant amount of time, since the maps can be large, with difficult to model walls.

CalebKoch:
2 points I think need a little discussion on how they will work:
1. If a player is in both sessions, and has access to a token, should that access move across? I would expect Default and the DM's access to copy over, but not sure on others.
2. Should the Location Setting list of sessions be in a new tab, or fall under Admin.

I've got some work started on this, but wanted to open this issue to get some discussion on these points.

Kruptein:
I know your feeling.

just for clarity's sake. Suppose you have a campaign running let's say curse of strahd.

and a couple of sessions in you start another CoS campaign and want to copy the start location.  You would be fine with it copying literally everything in that location then right?

Depending on how clean your map is, you might need to cleanup some things, closing doors that have been opened, removing tokens from players not associated with the new campaign, etc etc

All things considered this will be much less effort ofcourse compared to having to draw the walls etc from scratch, but just making sure if that's what you envision

CalebKoch:
Yup was definitely expecting that if I had opened doors and had tokens that had been defeated, those steps would need to be undone.

Kruptein:
> 1. If a player is in both sessions, and has access to a token, should that access move across? I would expect Default and the DM's access to copy over, but not sure on others.

Though call honestly, chances are pretty high that a person is going to play a different character I would say, but in that case you just remove the old token, so I don't think it would be bad to have the access copy over ?

> 2. Should the Location Setting list of sessions be in a new tab, or fall under Admin.

No particular preference yet, but it would be best if it was some button or other call to action that initiates a modal to choose from, so that we only need to fetch the session + location information when this modal is opened and not when the DM settings are opened for example

[...]

PR #706, opened 2021-03-14
Copy location

CalebKoch:
Intended to address #697 

Still have to make sure it copies over Aura's, trackers, labels, and groups. But wanted to send this out a little early to see if I could get any feedback. I feel like there should be a better way to do the make_copy methods, but I'm not super familiar with peewee, and I couldn't find anything in the documentation that would seem to help.

### DIFF
--- a/client/src/game/store.ts
+++ b/client/src/game/store.ts
@@ -346,7 +346,7 @@ class GameStore extends VuexModule implements GameState {
     cloneLocation(data: { location: number; room: string; sync: boolean }): void {
         const location = this.locations.find((l) => l.id === data.location);
         if (location === undefined) {
-            throw new Error("unknown location rename attempt");
+            throw new Error("unknown location clone attempt");
         }
         if (data.sync) sendLocationClone({ location: data.location, room: data.room });
     }
--- a/client/src/game/store.ts
+++ b/client/src/game/store.ts
@@ -15,6 +15,7 @@ import { toSnakeCase } from "../core/utils";
 import { sendClientLocationOptions, sendDefaultClientOptions, sendRoomClientOptions } from "./api/emits/client";
 import {
     sendLocationArchive,
+    sendLocationClone,
     sendLocationOrder,
     sendLocationRemove,
     sendLocationRename,
@@ -81,7 +82,7 @@ class GameStore extends VuexModule implements GameState {
     invertAlt = false;
     /**
      *  The desired size of a grid cell in pixels
-     
+
      *  !! This variable must only be used for UI purposes !!
      *  For core distance logic use DEFAULT_GRID_SIZE
      *  The zoom code will take care of proper conversions.
@@ -339,6 +340,15 @@ class GameStore extends VuexModule implements GameState {
         }
         location.archived = false;
         if (data.sync) sendLocationUnarchive(data.id);
+    }
+
+    @Mutation
+    cloneLocation(data: { location: number; room: string; sync: boolean }): void {
+        const location = this.locations.find((l) => l.id === data.location);
+        if (location === undefined) {
+            throw new Error("unknown location rename attempt");
+        }
+        if (data.sync) sendLocationClone({ location: data.location, room: data.room });
     }
 
     @Mutation
--- a/client/src/game/ui/settings/location/LocationAdminSettings.vue
+++ b/client/src/game/ui/settings/location/LocationAdminSettings.vue
@@ -5,18 +5,26 @@ import { Prop } from "vue-property-decorator";
 
 import InputCopyElement from "@/core/components/InputCopyElement.vue";
 import ConfirmDialog from "@/core/components/modals/ConfirmDialog.vue";
+import SelectionBox from "@/core/components/modals/SelectionBox.vue";
+import { baseAdjustedFetch } from "@/core/utils";
+import { RoomInfo } from "@/dashboard/types";
 import { gameStore } from "@/game/store";
+
+import i18n from "../../../../i18n";
 
 @Component({
     components: {
         ConfirmDialog,
         InputCopyElement,
+        SelectionBox,
     },
 })
 export default class LocationAdminSettings extends Vue {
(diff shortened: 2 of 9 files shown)


Kruptein:
Hey thanks for picking this up, I first need to do some hotfix work, I'll see if I can manage to get a review in today, otherwise I'll tackle this wednesday :)

CalebKoch:
Sounds good. 
I just got Aura's and trackers added. My initial search didn't return where I need to would update for Groups, and Label's seems structured a little different. I'm pretty busy the rest of the week, so I probably won't be able to dig into those last two issues untill next Monday, but I would still appreciate an early review.

CalebKoch:
Just added copying labels, and fixing Groups. Should be ready to go, whenever you have some time to review.

CalebKoch:
This pr should be good to go whenever you have a chance!

client/src/game/store.ts
```diff
@@ -341,6 +342,15 @@ class GameStore extends VuexModule implements GameState {
         if (data.sync) sendLocationUnarchive(data.id);
     }
 
+    @Mutation
+    cloneLocation(data: { location: number; room: string; sync: boolean }): void {
+        const location = this.locations.find((l) => l.id === data.location);
+        if (location === undefined) {
+            throw new Error("unknown location rename attempt");
```
Kruptein, line 349:
```suggestion
            throw new Error("unknown location clone attempt");
```

client/src/game/ui/settings/location/LocationAdminSettings.vue
```diff
@@ -53,12 +61,40 @@ export default class LocationAdminSettings extends Vue {
...
+
+            gameStore.cloneLocation({
+                location: this.location,
+                room: roomName,
+                sync: true,
+            });
+
+            this.$emit("close");
+        } else {
+            // TODO: Handle error codes. Don't see any type of Notification modal. Should it just console.log?
```
Kruptein, line 88: Keep it as a console log for now

server/api/socket/location.py
```diff
@@ -358,6 +367,36 @@ async def add_new_location(sid: str, location: str):
...
+        logger.warning(f"{pr.player.name} attempted to clone locations.")
+        return
+    try:
+        room = (
+            Room.select().where(
+                (Room.name == data["room"]) & (Room.creator == pr.player)
+            )[0]
+        )
+    except IndexError:
+        logger.warning(f"Destination room %s not found.", data["room"])
```
Kruptein, line 384:
```suggestion
        logger.warning(f"Destination room {data['room']} not found.")
```

server/models/shape/__init__.py
```diff
@@ -118,6 +119,52 @@ def center_at(self, x: int, y: int) -> None:
...
+            annotation=self.annotation,
+            draw_operator=self.draw_operator,
+            index=self.index,
+            options=self.options,
+            badge=self.badge,
+            show_badge=self.show_badge,
+            default_edit_access=self.default_edit_access,
+            default_vision_access=self.default_vision_access,
+            is_invisible=self.is_invisible,
+            default_movement_access=self.default_movement_access,
```
Kruptein, line 145: is_defeated is missing here

server/models/shape/__init__.py
```diff
@@ -118,6 +119,52 @@ def center_at(self, x: int, y: int) -> None:
...
+            is_locked=self.is_locked,
+            angle=self.angle,
+            stroke_width=self.stroke_width,
+            asset=self.asset,
+            group=None if self.group is None else self.group.make_copy(),
```
Kruptein, line 150: This creates a new group for each member. So for example a group with 3 shapes will result in 3 separate groups

server/models/shape/__init__.py
```diff
@@ -144,6 +196,11 @@ def __repr__(self):
     def as_dict(self):
         return model_to_dict(self, recurse=False, exclude=[Tracker.shape])
 
+    def make_copy(self, new_shape):
+        dict = self.as_dict()
+        dict['uuid'] = str(uuid4())
+        copy = type(self).create(shape=new_shape, **dict)
```
Kruptein, line 202:
```suggestion
        type(self).create(shape=new_shape, **dict)
```

server/models/shape/__init__.py
```diff
@@ -165,6 +222,11 @@ def __repr__(self):
     def as_dict(self):
         return model_to_dict(self, recurse=False, exclude=[Aura.shape])
 
+    def make_copy(self, new_shape):
+        dict = self.as_dict()
+        dict['uuid'] = str(uuid4())
+        copy = type(self).create(shape=new_shape, **dict)
```
Kruptein, line 228:
```suggestion
        type(self).create(shape=new_shape, **dict)
```

server/models/shape/__init__.py
```diff
@@ -212,6 +274,11 @@ def get_center_offset(self, x: int, y: int) -> Tuple[int, int]:
     def set_location(self, points: List[List[int]]) -> None:
         logger.error("Attempt to set location on shape without location info")
 
+    def make_copy(self, new_shape):
+        table = type(self)
+        dict = self.as_dict(exclude=[table.shape])
+        copy = table.create(shape=new_shape, **dict)
```
Kruptein, line 280:
```suggestion
        table.create(shape=new_shape, **dict)
```

Kruptein (changes requested):
Good job, some minor things and a bug with groups, but otherwise works great!

There is one special shape that will not properly be copied currently (ToggleComposite) because it has uuid references to other shapes that are not tracked, but I'm fine with leaving that out of scope

server/models/shape/__init__.py
Kruptein: This hasn't been addressed yet

CalebKoch: Yeah, had a couple busy weekends, and it took a bit to come up with a good way to make this work. Just pushed something up that takes care of it.

PR #713, opened 2021-03-19
by default, comment public_name config option

dthv:
having an empty configuration value can be confusing, it's better to comment it to show it is optional


Kruptein:
Add an entry to the Changed section of the changelog

dthv:
Done

PR #729, opened 2021-04-01
i18n: client/dm options to settings

### LINKED ISSUE #714: [BUG] Client Options vs Client Settings
This is just a minor one: The sidebar has a button called `Client Options` which opens a modal titled `Client Settings`.
This should be harmonised (esp. since the documentation, at the moment, introduces a third variant called `In-game Settings`).

dthv:
This closes #714
I changed DM 'options' to 'settings' as well to add to conformity. Changes were not only to the i18n strings but also to source to be consistent.

Not sure whether this is important enough to be mentioned within the changelog. Only pro I'd see for this would be to have an alert for translators, but translations themselves do not necessarily follow the conflict in English. So it might also be simply changed and that's it. There's still `git blame`. :)

### DIFF
--- a/client/src/game/ui/menu/MenuBar.vue
+++ b/client/src/game/ui/menu/MenuBar.vue
@@ -126,7 +126,7 @@ export default class MenuBar extends Vue {
                         <div v-if="!notes.length" v-t="'game.ui.menu.MenuBar.no_notes'"></div>
                     </div>
                 </div>
-                <!-- DM OPTIONS -->
+                <!-- DM SETTINGS -->
                 <button
                     class="menu-accordion"
                     @click="openDmSettings"
--- a/client/src/game/ui/menu/MenuBar.vue
+++ b/client/src/game/ui/menu/MenuBar.vue
@@ -127,7 +127,11 @@ export default class MenuBar extends Vue {
                     </div>
                 </div>
                 <!-- DM OPTIONS -->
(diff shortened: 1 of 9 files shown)


dthv:
PR_flow fail due to use of coding style not preferred by npm in src/game/ui/menu/MenuBar.vue:130:24 (kept from original)

Kruptein:
Yeah 'settings' is one char longer than 'options' and the linter wants to save the line split across multiple lines now, you can fix it manually or see if running `npm run lint` locally is able to autofix it.

client/src/game/ui/menu/MenuBar.vue
```diff
@@ -127,7 +127,7 @@ export default class MenuBar extends Vue {
                     </div>
                 </div>
                 <!-- DM OPTIONS -->
```
Kruptein, line 129:
```suggestion
                <!-- DM SETTINGS -->
```

dthv:
> you can fix it manually or see if running `npm run lint` locally is able to autofix it

So this *actually* needs fixing?
(sry, npm-n00b, here…)

Kruptein:
> > you can fix it manually or see if running `npm run lint` locally is able to autofix it
> 
> So this *actually* needs fixing?
> (sry, npm-n00b, here…)
> 
> 

Yeah :D, this is not really related to npm itself. Enforcing linter rules is one of the ways to make sure a codebase stays consistent in style while being worked on by multiple people

PR #730, opened 2021-04-12
Docker image building on ARM

jirsat:
This PR should make possible to build docker image on ARM machine. Changes are only in BUILDER, so final image size is the same as before.
I tested it on Raspberry PI 4B running Raspbian buster.


Kruptein:
Sorry for the long wait, just curious why is python3 needed in the build step? Is there some python dependency to build the client ?

jirsat:
Python is a requirement for node-gyp
https://github.com/nodejs/node-gyp#installation

Without this deasync cannot be installed.

I attached a part of  a log, if you're interested.
[build.log](https://github.com/Kruptein/PlanarAlly/files/6417143/build.log)

Kruptein:
Np, I believe you, was just curious

PR #759, opened 2021-05-14
feat(PlayerViewport): Add player viewport support

Kruptein:
This PR adds a new toggle in the DM/Admin settings to show player viewports.

When toggled on for a player, a new rectangle will appear on the DM layer that reflects the current viewport that player has.
You are able to move this rectangle and thus update the user's viewport as a DM.

This is especially useful for cases where you use a second monitor to show player-view.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,13 @@ These usually have no immediately visible impact on regular users
 -   Physical (mini) grid size
     -   Indicate grid size in function of mini dimensions and PPI
     -   Disables zoom tool when enabled
--   [tech] API Server is now disabled by default and can be enabled through the server_config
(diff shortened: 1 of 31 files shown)


dthv:
Should this correspond to a "allow viewport movement by DM" setting?

Kruptein:
We could, but I'm going to raise a counterpoint.

Moving a viewport is a very explicit action the DM needs to do, it's not something they can do by accident.  If a player is not comfortable with the DM controlling their view, it should be a very straightforward conversation were one asks please don't do that. If that's something the DM does not want to listen to, then I think you have other problems in your session :D
It's also not the only way the DM can move a player around. Even if a client option was available to prevent it, actions like "Bring Players" would still trigger a viewport change.

dthv:
I am convinced. :)

PR #806, opened 2021-07-25
feat(Dice): Further tweaks

Kruptein:
- Reduces the timer to automatically roll from 4 seconds to 1 second.
- Increases speed with which dice are thrown
- Randomize position and direction from where the dice are thrown
    - The dice will always be thrown from one of the sides vaguely towards the center
- Use the player's ruler colour as the colour for the dice
    - One small issue right now is that the font colour of the number is always black
- Increase wall height of dice box to prevent dice leaving the area
- Dice are cleared from the board when results are closed
- Limit height of history element with scroll behaviour
- Dice should stop rolling faster
- Add shadow to dice
- Add roof to make sure any funky dice throw stays within the box

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -12,7 +12,7 @@
                 "@fortawesome/free-regular-svg-icons": "^5.15.3",
                 "@fortawesome/free-solid-svg-icons": "^5.15.3",
                 "@fortawesome/vue-fontawesome": "^3.0.0-4",
-                "@planarally/dice": "^0.4.0-alpha.2",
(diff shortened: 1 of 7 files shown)


develroo:
Excellent job there. Nails 80% of the issues I had with die rolling. Now being able to show the results on every player's screen might be a good idea too. But not sure how that goes. It just has to be the result though, not the animation, I think. 

Cheers!

Kruptein:
Just the results is something I also would like to share to other players indeed!

Issue #672, opened 2021-02-12
[FEAT] Server Admin Console commands


Airforc:
Add commands to the server console for the server admin to manage basic functions:

/listaccts (list all player usernames/emails if supplied during registration, maybe list campaigns that accounts are attached to?)
/resetpswd <username > (resets account password to default 1234567890 or something in case of forgotten password)
/listcampaigns (list all created campaigns and useraccts that are assigned to dm/co-dm role)
/delacct <username> (delete user account, maybe needs some logic if useracct is dm of campaign?)
/delcamp <campaign> (deletes campaign)
/chgdm <campaign> <new dm acct> (replace current dm acct with new dm account on already created campaign)
/campassets <campaign> (returns information on space used for campaign assets, probably important for checking what campaign is using space)
/chgemail <username> (change usernames email account if email function is used)

dthv:
@Kruptein if you see this as a valuable feature, I could try to get up a basic python3 tool that does it. Yet, I'm afraid it would probably be outside of the server console, at least at first.

Think of something like `planarmanager list accs`, `planarmanager list campaigns` etc.

Kruptein:
go for it :)

Kruptein:
I personally would just use full command words btw `list accounts` with maybe a shorthand notation like `list accs` available too, but these commands are not something you have to do often and rapidly so I think it's better to write them out in full.

Airforc:
@Kruptein @dthv 

I have no preference, just something that allows easy maintenance of users/campaigns/assets, some of those commands probably could use additional parameters to specify which account to perform the command on.

I think a separate script is fine, but hopefully over time it gets merged into the main script :)

dthv:
Oh, I certainly plan (with the permission of our benevolent developer, of course :wink: ) to open a PR into this repository. With `outside of the server console` I meant that one would use a different shell than the one PA runs in.

Don't expect progress before March, though. Sorry.

Airforc:
@dthv honestly, if I brush up on my python, I might take a crack at it.

dthv:
Well, I was about to use this as a mini-project to brush up *my* python but that shall not hinder you. Feel free. :D

Kruptein:
Hey there, has any of the two of you been able to do something in this regard? :D

If nobody has the time which is understandable, I can pick it up myself as well

CalebKoch:
I assume after a few months without activity you weren't still working on this, so I took a stab at this for some of the commands in this issue. 

[...]

PR #831, opened 2021-10-04
admin console

CalebKoch:
Changes for part of #672 
Adds support for 

- Listing Users
- Listing Rooms
- Removing Users
- Removing Rooms
- Resetting Passwords


server/planarserver.py
```diff
@@ -129,16 +122,140 @@ async def start_servers():
     print()
 
 
-main_app.on_shutdown.append(on_shutdown)
+def server_main(_args):
```
Kruptein, line 125: This is seemingly not used

server/planarserver.py
```diff
@@ -129,16 +122,140 @@ async def start_servers():
...
+        )
+        sys.exit(1)
 
-if __name__ == "__main__":
+    mimetypes.init()
+    mimetypes.types_map[".js"] = "application/javascript; charset=utf-8"
+
+    save.check_save()
+
+    loop = asyncio.get_event_loop()
```
Kruptein, line 139: I would revert this back to the top of the file.
The `if sys.platform == "win"` depends on the presence of this variable and moving that block into this function is not so pretty in my opinion.

server/planarserver.py
```diff
@@ -129,16 +122,140 @@ async def start_servers():
...
+    password = args.password
+    if not password:
+        first_password = getpass.getpass()
+        second_password = getpass.getpass("Retype password:")
+        while first_password != second_password:
+            print("Passwords do not match.")
+            first_password = getpass.getpass()
+            second_password = getpass.getpass("Retype password:")
+        password = first_password
+    user = User.by_name(args.name)
```
Kruptein, line 183: Maybe fetch the user first before doing the password prompts.
If the user does not exist (e.g. typo) then typing the password twice was not in vain.

server/planarserver.py
Kruptein: This still needs to be addressed (I'm talking about the `_args` if that was not clear

CalebKoch: Sorry, Tried to respond to this, but been a while since I used GitHub's reviews. 

_args isn't used in this function, but it is going to be provided, by line 263, so it needs to be in the signature. Variable names starting with an `_` not being used is a convention used in many linters, so I figured that would make it clear. 
Any suggestions on how I could make it more clear that it is not being used, but needed? I could rename it to just `_` or `_unused`

CalebKoch: I'd have to add some manual parsing logic in the case that the script is called without a sub command, since right now it's just providing None, but that could be done. I'll try to make that update this evening, and see how that works.

Kruptein: Nah don't worry about it, if it's going to add a lot of extra logic, I'll just merge it as is.

CalebKoch:
> 
> 
> Hm I just realized that removal of rooms probably needs some extra work as a room name is not unique, multiple users could have a room with the same name.

Should be addressed with #836

PR #848, opened 2021-11-04
feat(Performance): Improve draw loop timings

Kruptein:
At the heart of PlanarAlly is the draw loop that rerenders the screen on changes in the scene.

This PR adds a couple of tricks to keep track of things that should be visible in the screen so that that specific calculation does not need to be done during the draw loop itself.

So at the cost of some more memory, the timings of the drawloop should go down, whether or not this is noticeable largely depends on your hardware (will be more noticeable on slower machines).

This is another set of changes that depends on correct invalidation of the "in view" set, which means that some cases might have been overlooked, let me know if rendering behaves unexpected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ These usually have no immediately visible impact on regular users
 ### Performance
 
 -   Cache Shape.points to prevent frequent recalculations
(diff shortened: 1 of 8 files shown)


Kruptein:
Another approach has been tested in my gameboard repository, which I'll be porting instead of this.

PR #858, opened 2021-12-13
bug(Map): Fix map-tool aspect ratio lock broken

Kruptein:
Due to a bug in the font-awesome vue dependency (which provides icons), the aspect lock toggle in the map tool no longer works.

I'm not sure exactly when it stopped breaking, but I added it to the changelog just in case it was already broken in the last release.

A workaround has been added to make it functional again.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -20,7 +20,7 @@
                 "socket.io-client": "^4.3.2",
                 "swiper": "^7.3.1",
                 "tinycolor2": "^1.4.2",
-                "vue": "^3.2.23",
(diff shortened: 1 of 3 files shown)


develroo:
When you say Aspect lock? Do you mean rescaling and then locking the map? Because year I noticed that was less accurate than it used to be, and I had to do it several times.

Kruptein:
I mean the icon you can toggle next to the x and y fields to maintain aspect ratio. So that if you modify the x value, the y value is modified accordingly to the aspect ratio

PR #866, opened 2022-01-26
feat(Logic): Add door logic

Kruptein:
This PR adds a first "logic" system that can be attached to a shape.

The door logic is a simple setting that allows the "door" to be opened or closed with a single click in play mode. This will simply toggle the 'blocksMovement' and 'blocksVision' settings of the specific shape, but 'wastes' less time.

When you hover the mouse cursor over a door, the cursor will change to a lock or unlock shape, which when left-clicked performs the action as described above.

Additionally it's possible to let players open specific doors as well, a set of permissions can be configured to decide which permissions players have. Permissions can be configured for all users with the 'default' label, or a specific user can get different permissions as well. Potential options are "enabled", "request" and "disabled", and defaults to "disabled" for the 'default' label.

With enabled, all relevant players can immediately toggle the state of a door on their own, whereas with request, they'll see the (un)lock icon when hovering over the door, but it will first send a request to the DM who can decide to approve this action or not.

Because this feature is intended to simplify things, and having to configure each door you draw as a PA door takes quite some time, it's also possible to immediately apply these settings in the draw tool so that you can draw all your doors in batches. You can find this under the new logic tab (cogwheels) in the draw tool.

In the future I would like to extend the permission system to also be able to add specific labels (as already used by the filter system and in the extra tab for shapes). This could allow the DM to create a `key` label which they can hand to a specific player once they receive the item in game, granting them access to open certain doors.

As explained earlier the cursor changes appearance when hovering over a door if you have either Enabled or Request permissions. **This means** that if you grant certain players these permissions they could in theory scout the presence of doors in the dark by just moving their mouse around and trying to find them as their cursor will change. There currently is no real function that can tell me whether the player has vision of a certain point, as that has only been relevant to the render process so far and that optimizes things making it unsuitable to quickly adapt. I have to think about how I want to fix this going forwards.


develroo:
Oh that is awesome! I have been manually adding door shapes to make sure the light does not leak around them and when you have portcullis it can be fiddly to add individual bars and then making them a group to move and remove them 

Gotta try this out.. Thanks @Kruptein

PR #872, opened 2022-02-16
tech(Systems): Experimenting with design changes

Kruptein:
This PR introduces some changes to some aspects of the internal PA design.

1) Local IDs

PA uses globally unique ids which are uuids (strings) to make sure that there can not be collisions when two people make objects at the same time, be it in the same session or completely different sessions. This is pretty standard stuff.

I wanted to experiment with using numerical ids in a local context (i.e. client side) to use for all client side references to these objects and only convert to global ids when crossing the network.

2) Explicitly typed IDs

Up until now the global ids, being strings were just typed as `string`. In typescript you can alias types but you can pass a string to some function expecting a particular alias of string fine, it doesn't restrict anything, just a convenience.

However you can do some tricks with 'brands', which I was already using for `GlobalPoint` and `LocalPoint`. Now there is also `GlobalId` and `LocalId` which improves readability and code quality. (already found 1 bug in this way!)

3) Split up data in conceptual chunks: "systems"

I've been reading a bunch about ECS recently, which fascinates me, but I didn't want to immediately restructure the entire PA codebase to use ECS. Instead I've taken to some of its principles and want to experiment with using specialized chunks that manage their own state instead of having the monolith that `Shape` has become which tracks all of its state.

My first experiment in this regard is moving the recently added logic systems to their own place in the codebase and so far I'm pretty happy with the result!


develroo:
Interesting. I am guessing you were [reading this](https://maxwellforbes.com/posts/typescript-ecs-what)  when thinking about it? 

I can see the advantages especially if it comes to exporting data like campaigns or maps or assets etc. 

Let's see how it works out. 

Cheers

PR #912, opened 2022-04-24
feat(Tooling): Add campaign exporter

Kruptein:
There already was export/import code hidden in the codebase, but I wasn't really happy with it.

This is a redo and instead of exporting to a .json file, this exports a .sqlite file as well as the assets (which the previous version didn't do).

This combination is all you need to just run the campaign (together with the main server code of course), instead of having to import it again.

Importing in an existing server (i.e. adding to an existing save file) has not yet been written, and will be for another release. For this reason I will probably not promote it as a new thing in the release notes, but will just be a tool for me to help debug other people's saves.

_Also there were a bunch of things missing in the previous exporter I noticed_

### DIFF
--- a/server/api/http/rooms.py
+++ b/server/api/http/rooms.py
@@ -1,7 +1,7 @@
+import asyncio
 from typing import Union
 from aiohttp import web
 from aiohttp_security import check_authorized
-from multidict import MultiDict
 from export.campaign import export_campaign
 
 from models import Location, LocationOptions, PlayerRoom, Room, User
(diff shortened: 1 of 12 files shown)


develroo:
Any clue about how to test that ? ;)

Kruptein:
Visit /api/rooms/<username>/<campaignname>/export.
This will start an export job on the server and once it's done you'll be able to access it at the url reported by your initial request.

10leej:
Did this make the 2022.1 release?

Kruptein:
It did, but because there is no import functionality yet, I didn't put it in the release notes.

You can technically already use the output as it is a valid save file and contains all the assets, you would just have to manually move them to the correct places.

PR #936, opened 2022-06-03
feat(Server): Restructure server files

Kruptein:
The root of the server folder contained both source code files (e.g. planarserver.py) as well as image files (e.g. static/assets/), build artifacts (e.g. static/vite/), config files (e.g. server_config.cfg) and save files (e.g. planar.sqlite).

The amount of files in the root folder was beginning to make it difficult to discern the actual source code files while working, and thus I decided to move all sourcecode to a `src` folder inside the server folder.

This does mean that starting the server will be slightly different now:
in the server folder: `python src/planarserver.py`
in the server/src folder: `python planarserver.py`


develroo:
So the Docker build file needs to copy to the new folder, correct ?

Kruptein:
> So the Docker build file needs to copy to the new folder, correct ?

I updated the Dockerfile to now start the python service from the new folder.
You shouldn't have to modify anything if everything went right.

PR #1142, opened 2023-01-11
chore: update Dockerfile to reduce the image size

tdurieux:
Hi there,

I've made a small improvement to the Dockerfile that I think could help optimize the image size. I did not update the changelog since it does not change the software itself. I hope it is what is expected.

Summary of the changes:
* I added `npm cache clean` after `npm install` to remove `npm` cache that is not needed inside the Docker image.
* I added the `--no-install-recommends` to with apt-get in order to not install unnecessary packages and reduce the image size.

Impact on the image size:
* Image size before repair: 342.55 MB
* Image size after repair: 311.48 MB
* Difference: 31.07 MB (9.07%)

I hope that you will find these changes useful to you. Let me know if you have any questions or concerns.

Thanks,

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,13 +10,13 @@ WORKDIR /usr/src/admin-client
 
 # Copy first package.json so changes in code dont require to reinstall all npm modules
 COPY admin-client/package.json admin-client/package-lock.json ./
-RUN npm i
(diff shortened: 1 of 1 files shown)


Kruptein:
Seems good! thanks for the contribution :)

Issue #1165, opened 2023-02-10
[BUG] 'Return to Token' buttons work through other HUD elements


rexy712:
**Describe the bug**
Not sure how to word this title. When a token goes offscreen, a little red popup with their icon inside appears at the edge of the screen. Clicking on that returns you to where that token is. The bug is that when the settings panel is open, this 'Return to Token' button can be clicked through the panel. This is especially annoying when trying to click on some of the buttons in the panel or selecting assets and your view is suddenly moved to that token.

This is also true of the location menu for the top part of the screen, the floor selection popup, all HUD buttons, 'Edit shape' menus, and the dice menu.

This does not affect the zoom bar in the top right or the quick edit/quick shape properties in the top right. Everything else I tested seemed to be affected.

**To Reproduce**
1. Create a shape and make it a token.
2. Scroll the grid so that the token is offscreen to the left, note where the 'Return to Token' popup is.
3. Open the settings panel.
4. Click the position where the 'Return to Token' popup was.
5. Instead of clicking on something in the settings panel, you are taken back to that token.

**Expected behavior**
I think that ideally the 'Return to Token' button would shift right when the settings panel is opened, so that it remains visible and functional. The other things (edit location panel, shape menu, floor selection popup, etc) should override the 'Return to Token' button rather than vice versa.

**Screenshots**

Video of testing as many HUD elements as I could think of.

https://user-images.githubusercontent.com/2442544/218175565-ff7b27bb-2777-47a6-9cf3-156c844e626c.mp4

**Desktop (please complete the following information):**
 - OS: gentoo linux
 - Browser: firefox 109.0.1, ungoogled-chromium 109.0.5414.119
 - Version: dev branch commit [06b7124](https://github.com/Kruptein/PlanarAlly/commit/06b712440ad38bec4fb32e514acfcbeeb3a45216)

PR #1175, opened 2023-02-11
bug(UI): Fix token direction UI rendering

rexy712:
A recent change from #1165 unintentionally caused the token direction UI to render as all red rather than displaying a preview of the token. This change reverts that side effect.

result of #1165:
 

after this fix:
image


Kruptein:
Haha oops that was sloppy from me :p I was using it for debugging reasons.

PR #1301, opened 2023-07-28
feat(Asset): Add asset sharing

### LINKED ISSUE #1298: [FEAT] Allow sharing assets between DMs
I've been the long term GM for my group but another member has recently started to host a game for us. I had set up a bunch of reusable assets with templates all set up that I figured they could use (eg goblins, kobolds, merchants, etc). We discovered that they can't see any of the assets I've uploaded and I can't see any they upload.

Not being able to see some assets makes sense, like hiding the appearance of a boss token for example. But common enemies or townsfolk would be nice to share.

Allow assets to be shared amongst multiple people, much like how tokens can be shared when in-game. Have edit and read permissions perhaps. It'd be a rather complex solution, but it would be really nice for the GMs to be able to share and not reduplicate work.

Kruptein:
This PR adds the capability to share assets with other users on the same server.
Either singular files or entire folders can be shared.

These are shared in either a "view only" or an "edit" mode. In the latter mode users can fully modify the contents of a folder including addition as well as removal of assets, so be warry with this right.

In view mode, the only thing the user can do is see the contents of the folder and drop the assets in their campaigns.

When a campaign is shared it appears in the root folder of the receiving user in the form of `${ownerName}-${assetName}` with a special icon signifying it was shared with you.  You're free to rename and move this asset around and it won't affect anything for the original user. This is purely a n anchor to link into the real asset when you enter it.

A warning banner is shown when you're browsing the content of a shared folder, indicating that if you do have edit rights, you need to be extra careful.

This closes #1298

### DIFF
--- a/client/src/apiTypes.ts
+++ b/client/src/apiTypes.ts
@@ -28,23 +28,24 @@ export interface ApiAssetShare {
   user: string;
   right: "view" | "edit";
 }
+export interface ApiAssetAdd {
+  asset: ApiAsset;
+  parent: AssetId;
+}
+export interface ApiAssetCreateFolder {
+  name: string;
+  parent: AssetId;
+}
+export interface ApiAssetCreateShare {
+  right: "view" | "edit";
+  user: string;
+  asset: AssetId;
+}
 export interface ApiAssetFolder {
   folder: ApiAsset;
   path: AssetId[] | null;
   sharedParent: ApiAsset | null;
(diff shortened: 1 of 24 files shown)


Kruptein:
Core logic is mostly done, the main two things remaining to do are:
- [ ] Update in-game assets sidebar to include shared assets
- [ ] Properly make the share modal UI

Issue #1324, opened 2023-11-25
[FEAT] A vision-blocking object should only block the area behind it, not itself


refparo:
**Is your feature request related to a problem? Please describe.**
When I set a shape to "block vision", it blocks not only the area away from the player character's line of sight, but also itself, making itself a black shape. For example, when a tree trunk is set to "block vision", I want to prevent my players seeing things after the trunk, but I also want to allow them to see the trunk itself, while PlanarAlly currently shows the trunk as a black circle.

**Describe the solution you'd like**
Players should be able to see the vision-blocking shapes.

**Describe alternatives you've considered**
I have tried to emulate this kind of effect by making the shape emit a light that covers only itself. However, this doesn't work when the shape is in the dimming area of a player's area of sight. (It doesn't dim along with the area)

**Additional context**
None.

Kruptein:
Hey, this is also something that has irked me in the past.

A naïve solution would be to just exclude themselves, however in the current way lighting is done, this means that the entirety of that object is ignored for vision. When reading this you might be confused as that probably sounds right. And this is fine for small objects, a rock or a tree trunk is indeed something you'll probably see in its entirety most of the time.

There are however 2 issues (that I experienced):
- Suppose you have some dungeon and you have decided to use thick rectangles to block out the walls. If a player walks past the wall they'll see these 30ft thick walls and then suddenly there is a thin wall, without doors. Our secret hidden room has been spoiled! (note: In practice I personally use the polygon tool for walls so it wouldn't really apply to me)
- A side-effect of the current lighting/vision implementation I wasn't able to resolve when I did a stab at this a couple of months ago, was that it wasn't possible to limit what you saw of vision-blocking items to just that which fell within a light aura. Instead if your vision covered part of the object, you saw the entire object.

Now with this monologue out of the way, there is a cheap solution to offer what you're looking for: Make it a new optional toggle per shape. In that case you could mark your tree trunk or a door as some new property which would exclude itself and reveal its entirety, which for small shapes should be fine.

It's just a bit more work on the DM's end to toggle this.

I would like to tackle this more properly at one point, especially (2) as for (1) I actually don't know if there is a proper solution.

Would this toggle be an acceptable solution for you ?

develroo:
[...]

PR #1338, opened 2023-12-11
feat(Vision): Exempt vision-blocking shapes from themselves

Kruptein:
This PR is an attempt to tackle #1324.

## Goal

With this implementation _all_ vision blocking shapes that are _closed_ will be visible themselves, but still block everything behind them.

This fixes a longstanding issue where you can't see the pillar/tree/whatever that is blocking the vision, instead your vision is just cutoff. As a DM you could try to give some glimpse by making the vision blocking shape slightly smaller, but this is an annoying balancing act.

## What are closed shapes

It's important to focus on the _closed_ aspect. What I mean with this is shapes that actually fill their content.

Most basic shapes fit this criterion (i.e. rectangles, circles, assets).

Some shapes do not fit this criterion (e.g. Text, Brush tool, and some shapes you can't manually draw (e.g. ruler lines etc), these shapes tend to not be interesting for vision blocking anyway.

That leaves us with the polygon tool. This has always had an explicit "closed polygon" toggle that dictates whether the content should be filled and also automatically closes the polygon if you didn't do so yourself. The polygon tool's behaviour thus will depend on its closed status.

## Why only closed shapes

This is simply because I need the fill for how I ended up implementing it. Open shapes can be too wild and unpredicting to make a guess that it's better to leave them out entirely.

As said earlier, most open shapes are not interesting to begin with. Only the open polygon tool is one I use a lot myself to draw walls.

The result is that open polygons will still behave as they did before, they form a hard-block of vision and the nice thing is, this gives us an alternative for when we explicitly don't want the self-excluding behaviour.

## What about the issues mentioned in #1324

I posed 2 problems in the linked issue I had a year ago when I last attempted to solve the same problem.

(1) Although excluding vision-blocking shapes from themselves is nice for small objects like pillars etc, it can potentially have drawbacks when used for more complext things. E.g. if all walls are drawn with rectangles that now can be looked into up to the other end of the wall, potential secret rooms can be spoiled by the notion that the wall suddenly becomes smaller

This still to a certain degree is an 'issue'. However:
- I personally always draw my walls with open polygons, given that they are not impacted, my secret rooms will actually not be spoiled at all
- I'm still pondering making this system either opt-in/opt-out for individual shapes instead of hard applying it to everything. In this case it's just an extra configuration thing and you set a default that makes most sense to you.

(2) I had some version working in the past, but it had the downside that if a part of a vision-blocking shape was in vision, the entire shape would become visible even if your light aura didn't reach all the way.

This implementation no longer has this issue :D

## Todo

Main todos are:
- think about opt-in/opt-out
- verify behaviour when things move, I'm sure I need to do some invalidation still
- do a check on performance impact

## Example

Example player vision:
image

FOW behind the scenes, notice the pillar is a full circle, the door is a full rectangle, but the walls are _open_ polygons.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ tech changes will usually be stripped from release notes for the public
 
 ### Changed
 
+-   Vision blocking shapes will now ignore themselves if they are closed
+    -   e.g. a tree trunk will be visible, but what's behind the tree trunk will remain hidden
+    -   Open polygons will behave as they have in the past
 -   [tech] ModalStack now supports dynamically inserted
 
 ### Fixed
--- a/client/src/apiTypes.ts
+++ b/client/src/apiTypes.ts
@@ -6,6 +6,7 @@ import type { AuraId } from "./game/systems/auras/models";
 import type { CharacterId } from "./game/systems/characters/models";
 import type { ClientId } from "./game/systems/client/models";
 import type { PlayerId } from "./game/systems/players/models";
+import type { VisionBlock } from "./game/systems/properties/types";
 import type { TrackerId } from "./game/systems/trackers/models";
 
 export type ApiShape = ApiAssetRectShape | ApiRectShape | ApiCircleShape | ApiCircularTokenShape | ApiPolygonShape | ApiTextShape | ApiLineShape | ApiToggleCompositeShape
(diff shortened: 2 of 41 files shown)


Kruptein:
The choice was made to make the behaviour configurable per shape.
Vision blocking now essentially is either: "no blocking", "complete blocking" (old behaviour), "behind blocking" (new behaviour).

This just gives more power to the end-user and doesn't suddenly surprise a bunch of users that update and suddenly have all their light/vision behaving differently.

There is only 1 location in the entire codebase were a decision is currently made for the user:
- Door toggles will toggle between Behind and None mode. Given the small surface doors usually have this seems like a safe call and it's one of the places were I personally felt the need to have Behind mode. If this ends up being a problem for people I can accomodate.

PR #1350, opened 2024-01-15
feat(DiceTool): Resolve #1327, #1328

### LINKED ISSUE #1327: [FEAT] Explain the dice rolls in the log
The dice log is too terse. If someone rolls 2d6, it does not explain what the rolls were, only what the sum is.

I'm asking for a detailed log, like `2d6 = 1+6`

### LINKED ISSUE #1328: [FEAT] Show who rolled what in the log
The dice log is too terse and doesn't show who rolled what without hovering over each roll. I constantly need this info

Example: `[user]: 1d6`

SuikaXhq:
Resolves #1327
Resolves #1328 

Note that in `client/src/game/dice/state.ts`, I just use the first element of result list because I don't quite understand why the list always contains only one result item. Using the first item just works well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ tech changes will usually be stripped from release notes for the public
     -   If the text area of a note is still in focus after 5 seconds and an edit was made, a server save is done
     -   _see the release notes for all the changes_
 -   [tech] ModalStack now supports dynamically inserted modals
+-   Dice history now contains user and details
 
 ### Fixed
 
--- a/client/src/apiTypes.ts
+++ b/client/src/apiTypes.ts
@@ -451,7 +451,7 @@ export interface Viewport {
 export interface DiceRollResult {
   player: string;
   roll: string;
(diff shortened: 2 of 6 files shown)


Kruptein:
> Note that in client/src/game/dice/state.ts, I just use the first element of result list because I don't quite understand why the list always contains only one result item. Using the first item just works well.

The underlying dice API offers capabilities to do multiple rolls at the same time that should not be added together (e.g. 2d6+3 1d4-2, would do the 2 rolls at the same time and keep track of their underlying totals and subtotals as 2 separate entries).

The frontend UI of PA does not offer this option however so it's fine to take the first element only.

Kruptein:
did you btw end up receiving my response to your mail?

SuikaXhq:
are you referring to the email I sent to you on Oct. 17? It seems failed to receive your response as I haven't found any email related to that since then.

Kruptein:
Oh, I'll try again then :V because I did reply to that mail

PR #1351, opened 2024-01-16
fix(TrackerSettings): Show scrollbar for long contents

SuikaXhq:
I make the tracker tab have a max height of 60vh and add scrollbar to it. Now long list of trackers is easier to access.


Kruptein:
Looks good, just missing an entry in the changelog!

SuikaXhq:
updated changelog for both PRs just now. I forgot it.

PR #1352, opened 2024-01-16
feat(SelectionInfo): Quick toggle or rotate for auras

SuikaXhq:
Replace the original quick edition of aura values with quick toggling and rotating.
RotationSlider is only shown when angle<360.

The toggling function is copied from TrackerSetting.vue. I thought it would be better if there is a unified function to access these states (also useful for mod API). It works at present though.

btw, I noticed that the orbit of RotationSlider's "slider" is actually larger than the rendered one. would that be a bug?

### DIFF
--- a/client/src/game/ui/SelectionInfo.vue
+++ b/client/src/game/ui/SelectionInfo.vue
@@ -3,12 +3,12 @@ import { computed, ref, type DeepReadonly } from "vue";
 import { useI18n } from "vue-i18n";
 
 import RotationSlider from "../../core/components/RotationSlider.vue";
-import { NO_SYNC, SERVER_SYNC } from "../../core/models/types";
+import { SERVER_SYNC } from "../../core/models/types";
 import { activeShapeStore } from "../../store/activeShape";
 import { getShape } from "../id";
 import { accessState } from "../systems/access/state";
 import { auraSystem } from "../systems/auras";
-import type { Aura, AuraId, UiAura } from "../systems/auras/models";
+import type { Aura, AuraId } from "../systems/auras/models";
 import { noteState } from "../systems/notes/state";
 import { type ClientNote, NoteManagerMode } from "../systems/notes/types";
 import { editNote, openNoteManager, popoutNote } from "../systems/notes/ui";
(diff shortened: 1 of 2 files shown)


client/src/game/ui/SelectionInfo.vue
```diff
@@ -69,6 +70,15 @@ function setValue(data: { solution: number; relativeMode: boolean }): void {
     }
 }
 
+function updateAura(aura: DeepReadonly<UiAura>, delta: Partial<Aura>, syncTo = true): void {
+    if (delta.value !== undefined && (isNaN(delta.value) || delta.value < 0)) delta.value = 0;
+    if (delta.dim !== undefined && (isNaN(delta.dim) || delta.dim < 0)) delta.dim = 0;
+    if (aura.temporary) {
+        auraSystem.add(aura.shape, { ...aura }, SERVER_SYNC);
+    }
```
Kruptein, line 78: I actually don't think you need to worry about these things. With the change you made, we're no longer modifying the `.value` or `.dim` and the temporary auras don't show up in the selection info either.

Kruptein:
> The toggling function is copied from TrackerSetting.vue. I thought it would be better if there is a unified function to access these states (also useful for mod API). It works at present though.

I made a small comment on that, I understand you just copied it, but for the use here you can just call the auraSystem directly. The original code is specifically to handle a bunch of conditions when creating new auras.

> btw, I noticed that the orbit of RotationSlider's "slider" is actually larger than the rendered one. would that be a bug?

Yeah this is a bug, I'll fix this, I also noticed that the two sliders are not in sync (the quick toggle slider and the one in the full tracker settings), but that's also on me, I've fixed that locally already, but I'm gonna wait for this one to land first.

Kruptein (changes requested):

client/src/game/ui/SelectionInfo.vue
SuikaXhq: all right, I will remove that.
btw, what does "NO_SYNC" do actually? It seems to keep sync anyway.

Kruptein: Yeah so while the mouse is still pressed, the changes are only applied locally. When the mouse is released, it's properly synced to the server and all other players.

PR #1358, opened 2024-01-21
feat(Notes): Change icons used for popouts

Kruptein:
The icon used for note popouts has been changed, as I found the previous version to imply too much that a completely new window would open (which is something that I want to add later).

The quick info icon for opening the note manager has also changed to a cog wheel, and the order of the two icons has changed.

Additionally the [share with players] action in the note dialog is removed as it did nothing and it feels somewhat out of place for that view.


develroo:
Yup, makes sense and looks good. I do like the idea of a pop out window too, though, as then it can be dragged on to a different desktop and allow you to arrange notes so they are one glance usable.

Screenshot from 2024-01-21 11-19-47
Screenshot from 2024-01-21 11-13-58

Kruptein:
Yeah, that's something I plan to do, but might postpone to a later release as to not delay the current one too much.

PR #1372, opened 2024-01-28
feat(Ruler): Add grid mode

Kruptein:
_WIP_

This PR adds an extra option called 'grid mode' to the ruler tool (and as a result the select tool as well when it's in ruler mode).

In grid mode the ruler calculates the distance between 2 points in terms of the number of grid cells it has to traverse (e.g. the number of squares in a square grid setup).

This results in something like:
image

Showing a total of 21 cells traversed which equals 105ft (with 1 grid = 5ft settings).


Kruptein:
image
Also supports hexes!

PR #1396, opened 2024-02-14
Feat(Ruler): Allow grid unit steps less than 0. Also fixes incorrect counting of cells in grid mode.

rexy712:
I realized that PA doesn't allow decimal numbers as a grid unit size while trying to make a regional map for use in an army combat scenario. I wanted to make hexes each 0.1 miles, and the UI appeared to let me, but it was silently being rejected.

I thought that it wouldn't hurt anything to allow such grid measurements, even if not very applicable to most situations. I feel like with the hex grid now working much smoother, it allows for more regional or overworld style map usage, at least in games like Pathfinder.

While I was working on this, I needed to make the 'Grid Mode' of the ruler tool function and noticed a couple bugs in the distance shown on screen. I fixed those issues while I was in working on that file since they were rather straightforward. Those being
- Measuring from a starting cell and remaining in that cell showed a distance of -1 cells
- Measurements in Grid Size Units were always 1 increment greater than they should have been.
- Using multiple rulers would give incorrect distances

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ tech changes will usually be stripped from release notes for the public
     -   Error toasts no longer have a timeout
     -   (Shape)Labels are no longer exported/imported
 -   [tech] Select tool: only take shapes in view into account
+-   Ruler tool: allows Unit Size less than 1.0
 
 ### Fixed
 
@@ -42,6 +43,8 @@ tech changes will usually be stripped from release notes for the public
         -   This should now more properly snap shapes that are larger than 1x1
 -   Ruler Tool:
     -   Snap now properly works for hex grids
+    -   Grid mode now starts count from 0 instead of -1
+    -   Grid mode correctly counts total cells when using multiple rulers
 -   Map Tool:
     -   Now better supports hex grids
 -   Spell tool: selecting another tool would swap to the Select tool instead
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -42,6 +43,8 @@ tech changes will usually be stripped from release notes for the public
         -   This should now more properly snap shapes that are larger than 1x1
 -   Ruler Tool:
     -   Snap now properly works for hex grids
+    -   Grid mode now starts count from 0 instead of -1
+    -   Grid mode correctly counts total cells when using multiple rulers
```
Kruptein, line 47: As grid mode is new, any fixes to it should not be included in the changelog as they never existed compared to the previous version.

Kruptein (changes requested):
A very small remark to fix, otherwise nice job!

Kruptein:
Ah you should also mention the QoL change to the spell tool in the changelog!

Ideally that was a separate PR, but I can ignore that for once ;)

rexy712:
> Ah you should also mention the QoL change to the spell tool in the changelog!
> 
> Ideally that was a separate PR, but I can ignore that for once ;)

I kinda thought of it as a part of this work since the  spell tool scales to fit the grid unit size, which is why I included it here.

It does seem like a separate issue that could have been done as a separate pull request. I'll try to better separate next time!

PR #1404, opened 2024-02-20
Fixed query when checking for valid Shape on initiative change

CalebKoch:
Fixed a query in sockets.initiative.

PeeWee allows just providing an id to it's query methods, such as get_or_none, but that requires the id to be an int. Since the Shape table has guid/string ids, the single argument was instead being passed as the entirety of the WHERE clause in the sql query.

This caused Sqlite to cast it to an int by only any digits at the start of the id. The result was if the id started with a non-zero digit, the query evaluated to True, and would return a shape, allowing the initiative to be updated. But if the id started with a letter, it would evaluate to false, so the query would return no results, which was interpreted as an invalid Shape. 

In searching the rest of the code base, it does seem like the other queries are correctly using keyword args, so this appears to be the only location with this specific issue.


Kruptein:
Good catch!

Issue #1406, opened 2024-02-21
[BUG] Note icon on shape can be drawn over by its parent shape


rexy712:
**Describe the bug**
Looking into the code to improve my implementation of the invisibility marker, I saw that note icons are currently drawn as unique shapes on the same layer as the base shape they attach to. So I thought that they could be drawn over by using the base shape's  'Move to front' and 'Move to back' context menu items, and sure enough they can.

The note icon can also be drawn over other shapes that are in the foreground compared to the base shape, but behind the note icon's shape. I believe that if a shape is in the foreground compared to the base shape, it should also always be in front of the note icon.

**To Reproduce**
1. Create a rectangle.
2. Add a note and mark the note to show icon on shapes.
3. Move the rectangle to front using the context menu.
4. Note icon is covered by the rectangle.

**Expected behavior**
The note icon should always be in the foreground compared to the shape it is attached to.

**Screenshots**

https://github.com/Kruptein/PlanarAlly/assets/2442544/72891b40-8370-40ed-89a2-24b000501390

**Desktop:**
 - OS: gentoo linux
 - Browser: ungoogled-chromium 121.0.6167.160, firefox 122.0.1
 - Version: dev branch commit [a44fd4d](https://github.com/Kruptein/PlanarAlly/commit/a44fd4d95920a5b98146a58cd1ddaf916d7a4ad0)

Kruptein:
Ah good catch, this one I wasn't aware of yet, but makes sense.

PR #1409, opened 2024-02-24
Set note icons into their own category of shape for rendering

rexy712:
Note icons are currently drawn as an ordinary shape which works in most ways. Since they are lumped in with all other shapes in a given layer, there exist some problems such as #1406 when it comes to rendering priority. The shape could become obscured by the shape it is supposed to be attached to.

This PR introduces a new class of shapes in a layer called dependent shapes. These are treated as shapes that depend on another 'parent' shape to exist and are therefore always drawn on top of the parent shape but still remain underneath other overlapping shapes as one would expect. This new class can be utilized to allow drawing of other icons such as an invisibility indicator in the future.

This solves #1406.

### DIFF
--- a/client/src/game/api/events/shape/core.ts
+++ b/client/src/game/api/events/shape/core.ts
@@ -32,9 +32,7 @@ socket.on("Shape.Set", (data: ApiShapeWithLayerInfo) => {
     let deps = undefined;
     if (old) {
         deps = old.dependentShapes;
-        for (const dep of deps) {
-            old?.removeDependentShape(dep.id, { dropShapeId: false });
-        }
+        old.removeDependentShapes({ dropShapeId: false });
         old.layer?.removeShape(old, { sync: SyncMode.NO_SYNC, recalculate: true, dropShapeId: true });
     }
     const shape = addShape(apiShape, floor, layer, SyncMode.NO_SYNC, deps);
--- a/client/src/game/api/events/shape/core.ts
+++ b/client/src/game/api/events/shape/core.ts
@@ -29,8 +29,15 @@ socket.on("Shape.Set", (data: ApiShapeWithLayerInfo) => {
     const old = getShapeFromGlobal(uuid);
     const isActive = activeShapeStore.state.id === getLocalId(uuid);
     const hasEditDialogOpen = isActive && activeShapeStore.state.showEditDialog;
-    if (old) old.layer?.removeShape(old, { sync: SyncMode.NO_SYNC, recalculate: true, dropShapeId: true });
-    const shape = addShape(apiShape, floor, layer, SyncMode.NO_SYNC);
+    let deps = undefined;
+    if (old) {
+      deps = old.layer?.getDependentShapes(old.id) ?? [];
+      for (const dep of deps) {
+        old.layer!.removeDependentShape(old.id, dep.id, { dropShapeId: false });
+      }
+      old.layer?.removeShape(old, { sync: SyncMode.NO_SYNC, recalculate: true, dropShapeId: true });
+    }
+    const shape = addShape(apiShape, floor, layer, SyncMode.NO_SYNC, deps);
 
     if (shape && isActive) {
         selectedSystem.push(shape.id);
--- a/client/src/game/api/events/shape/core.ts
+++ b/client/src/game/api/events/shape/core.ts
@@ -31,11 +31,11 @@ socket.on("Shape.Set", (data: ApiShapeWithLayerInfo) => {
     const hasEditDialogOpen = isActive && activeShapeStore.state.showEditDialog;
     let deps = undefined;
     if (old) {
-      deps = old.layer?.getDependentShapes(old.id) ?? [];
-      for (const dep of deps) {
-        old.layer!.removeDependentShape(old.id, dep.id, { dropShapeId: false });
-      }
-      old.layer?.removeShape(old, { sync: SyncMode.NO_SYNC, recalculate: true, dropShapeId: true });
+        deps = old.dependentShapes;
+        for (const dep of deps) {
+            old?.removeDependentShape(dep.id, { dropShapeId: false });
+        }
+        old.layer?.removeShape(old, { sync: SyncMode.NO_SYNC, recalculate: true, dropShapeId: true });
     }
     const shape = addShape(apiShape, floor, layer, SyncMode.NO_SYNC, deps);
 
--- a/client/src/game/shapes/shape.ts
+++ b/client/src/game/shapes/shape.ts
@@ -13,7 +13,7 @@ import {
 } from "../../core/grid";
 import { rotateAroundPoint } from "../../core/math";
 import { mostReadable } from "../../core/utils";
-import { generateLocalId, getGlobalId, getShape } from "../id";
+import { generateLocalId, getGlobalId, getShape, dropId } from "../id";
 import type { GlobalId, LocalId } from "../id";
 import type { ILayer } from "../interfaces/layer";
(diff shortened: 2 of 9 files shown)


Kruptein:
I did a quick glance through the code and you seem to have thought of plenty of cases, which is nice.

Is there a particular reason why you chose to track the dependents on layer level instead of on the shapes themselves? Because I feel like some logic would not be necessary in that case (e.g. layer changes). It's not a dealbreaker for me, but something I was curious about.

I have a busy day, but I'll try to give this a proper review tomorrow.

rexy712:
I had just thought it fit better with how the data was stored for everything else. It does make it easier to put the dependents in the shape itself. I can do that without too much work from what I have now and simplify the logic in some places.

I'll wait for your review before changing it.

client/src/game/temp.ts
```diff
@@ -90,7 +102,20 @@ export function addShape(shape: ApiShape, floor: string, layerName: LayerName, s
     if (sh === undefined) {
         return;
     }
+
+    // if this shape has been moved to a new location, notes are already set up for it client-side.
+    // we need to reattach the notes to this new shape since the local id is changed!
+    const noteList = Array.from(noteState.mutableReactive.notes.values());
```
Kruptein, line 108: I don't think this needs mutable access, `noteState.reactive.notes.values())` should suffice.

client/src/game/temp.ts
```diff
@@ -90,7 +102,20 @@ export function addShape(shape: ApiShape, floor: string, layerName: LayerName, s
     if (sh === undefined) {
         return;
     }
+
+    // if this shape has been moved to a new location, notes are already set up for it client-side.
```
Kruptein, line 106: Although this does work, I actually don't think this is the best place to tackle this.

Ideally the note system handles this internally. To do this it should first become a `ShapeSystem` instead of a normal `System`, in which case it would be set up to receive information when a shape is added/removed.

This can however be done in a separate PR, so it's not blocking for me.

Kruptein (changes requested):
As mentioned earlier, I would move the dependenShapes to Shape itself as Layer only ever interacts with it in function of a specific shape and not really in function of itself.

One small nitpick is that I prefer to have an empty line between functions, so if you could fix that in a couple of places as well that would be nice :D

client/src/game/temp.ts
rexy712: When trying to implement this I did originally use noteState.reactive but found that when passing `note` to `noteSystem.hookupShape(note, sh.id)` a TS2345 error was produced stating that readonly type is not assignable to a mutable type. So I was forced to use mutableReactive.

<details>
<summary>The full error</summary>

```
src/game/temp.ts:103:32 - error TS2345: Argument of type '{ readonly location: number | null; readonly text: string; readonly title: string; readonly creator: string; readonly uuid: string; readonly shapes: readonly GlobalId[]; readonly showOnHover: boolean; readonly showIconOnShape: boolean; readonly isRoomNote: boolean; readonly access: readonly { ...; }[]; readonly tags...' is not assignable to parameter of type 'ClientNote'.
  Type '{ readonly location: number | null; readonly text: string; readonly title: string; readonly creator: string; readonly uuid: string; readonly shapes: readonly GlobalId[]; readonly showOnHover: boolean; readonly showIconOnShape: boolean; readonly isRoomNote: boolean; readonly access: readonly { ...; }[]; readonly tags...' is not assignable to type 'Omit<ApiNote, "tags">'.
    Types of property 'shapes' are incompatible.
      The type 'readonly GlobalId[]' is 'readonly' and cannot be assigned to the mutable type 'GlobalId[]'.

103         noteSystem.hookupShape(note, sh.id);

```

</details>

Kruptein: True, that would indeed error with the existing code; `hookupShape` however does not need a mutable reference to note for it to function, so its signature can be either modified to accept a `DeepReeadonly<ClientNote>` instead or what's probably better, only expect a note ID instead of a full note, as it only uses `note.uuid` in its code.

Could you update that?

client/src/game/shapes/shape.ts
```diff
@@ -699,4 +715,38 @@ export abstract class Shape implements IShape {
...
+
+    removeDependentShapes(options: {dropShapeId: boolean}): void {
+        if (options.dropShapeId) {
+            for (const shape of this.dependentShapes) {
+                dropId(shape.id);
+            }
+        }
+        while (this._dependentShapes.length) {
+            this._dependentShapes.pop();
+        }
```
Kruptein, line 747: This can just be `this._dependentShapes = []`

client/src/game/api/events/shape/core.ts
```diff
@@ -29,8 +29,15 @@ socket.on("Shape.Set", (data: ApiShapeWithLayerInfo) => {
...
     const isActive = activeShapeStore.state.id === getLocalId(uuid);
     const hasEditDialogOpen = isActive && activeShapeStore.state.showEditDialog;
-    if (old) old.layer?.removeShape(old, { sync: SyncMode.NO_SYNC, recalculate: true, dropShapeId: true });
-    const shape = addShape(apiShape, floor, layer, SyncMode.NO_SYNC);
+    let deps = undefined;
+    if (old) {
+        deps = old.dependentShapes;
+        for (const dep of deps) {
+            old?.removeDependentShape(dep.id, { dropShapeId: false });
+        }
```
Kruptein, line 37: I believe this can be simplified to the following ?

```suggestion
        old.removeDependentShapes({ dropShapeId: false });
```

If not, the `removeDependentShapes` function is always called with `dropShapeId` true, in which case the argument can just be removed alltogether.

client/src/game/shapes/shape.ts
```diff
@@ -699,4 +715,38 @@ export abstract class Shape implements IShape {
         }
         return false;
     }
+
+    // DEPENDENT SHAPES
+
+    addDependentShape(shape: IShape): void {
+        if (this.floorId === undefined || this.layerName === undefined) return;
+
+        shape.setLayer(this?.floorId, this?.layerName);
```
Kruptein, line 724: the `Shape.setLayer` function itself should probably also update its dependents right? (I can't tag the line itself because it's not part of the PR apparently)

client/src/game/shapes/utils.ts
```diff
@@ -189,10 +191,14 @@ export function deleteShapes(shapes: readonly IShape[], sync: SyncMode): void {
         const sel = shapes[i]!;
         if (sync !== SyncMode.NO_SYNC && !accessSystem.hasAccessTo(sel.id, false, { edit: true })) continue;
         const gId = getGlobalId(sel.id);
-        if (gId) removed.push(gId);
+        if (gId) {
+            removed.push(gId);
+            noteSystem.unhookShape(sel.id);
```
Kruptein, line 196: In the future these fixes to other things, should be done in separate PRs ideally.

Kruptein (changes requested):
I did another pass with some small commments

client/src/game/api/events/shape/core.ts
rexy712: Ah my bad. I didn't originally have the `removeDependentShapes` function just `removeDependentShape` and decided to add the former for things like this, but forgot to change this old code.

client/src/game/shapes/utils.ts
rexy712: Yeah, sorry. I can revert and include it as a separate PR if you want.

client/src/game/shapes/shape.ts
rexy712: Yeah, good call.

Issue #1410, opened 2024-02-24
[BUG] Adding or removing shapes from notes not being received by clients until a refresh


rexy712:
**Describe the bug**
When adding a shape to a note, your client sees the update happen immediately. Other clients do not see the change until they refresh the page or until view access for the note is removed and granted again.

The event is being sent to the server, but it seems like the other clients never receive the update from the server.

**To Reproduce**
1. Create a shape.
2. Create a note, give default view access. They can see the note in the notes manager.
3. In the note manager, select 'shapes' submenu for the note.
4. Add the shape from step 1 to the note. Selecting the shape as this player shows 1 note in the Selection View.
5. As second player: Select the shape. In the Selection View, it still says 0 notes.
6. As second player: Refresh the page, reselect shape. Selection View now says 1 note.

**Expected behavior**
The status of the notes should update immediately.

**Screenshots**

This video shows that the icons are not updated, but it is also evident when selecting the shape. This is not an icon bug, but rather the client is not receiving the shape add event.

https://github.com/Kruptein/PlanarAlly/assets/2442544/8c0236a0-dfe7-4403-b847-a01f6a4c771e

**Desktop:**
 - OS: gentoo linux
 - Browser: ungoogled-chromium 121.0.6167.160, firefox 122.0.1
 - Version: dev branch commit [a44fd4d](https://github.com/Kruptein/PlanarAlly/commit/a44fd4d95920a5b98146a58cd1ddaf916d7a4ad0)

PR #1415, opened 2024-02-25
bug(Notes): Implement client handling of Note.Shape.Remove events

rexy712:
In #1413, the Note.Shape.Add event was properly handled so that clients saw the changes immediately. I noticed that Note.Shape.Remove events were still not being received properly.

Turns out it wasn't implemented client-side. This PR implements handling of these events. Finishes closing #1410


Kruptein:
Woops

PR #1448, opened 2024-06-30
bug(Chat): Do url image check serverside to prevent client CORS issues

Kruptein:
The feature that auto-detects images in chat was doing the image check from the client side, which causes edge cases where images that are for some reason CORS restricted to not load.

The check has been moved to the server, for which CORS does not apply.


develroo:
Can you give a working image URL for this because I don't actually see a change in all the ones I shared before? Just in case there is an edge case when running in a container.

develroo:
Hmmm, Nope still not working for me. This is the url I am pasting into chat. 

`https://www.pngall.com/wp-content/uploads/5/Disney-Princess-Aurora.png`

Is there any way I can debug what is going on ? I am not seeing 


FWIW the server config is set as thus:

```
# Allowed CORS origins
#     If not specified, only the host runnig the server is allowed
#     Allowed values are: ['*'], ['some-address.xyz', 'other-address.xyz'], []
#     More information on the meaning of these values can be found at
#     https://python-socketio.readthedocs.io/en/latest/api.html#asyncserver-class
cors_allowed_origins = *
```

Kruptein:
> > Does the server have access to the outside world?
> 
> Yes. As I explained before, using the `!` `[]` `()` notation on the same URL works.

This is not really relevant, as the `` syntax is purely client side, whereas this feature happens server side.

> 
> Hmmm, Nope still not working for me. This is the url I am pasting into chat.
> 
> `https://www.pngall.com/wp-content/uploads/5/Disney-Princess-Aurora.png`

I tested with this url, but I can confirm that it no longer works on my latest build. I must have changed something last minute before pushing without testing.

> Is there any way I can debug what is going on ? I am not seeing any logs?
> 
> FWIW the server config is set as thus:
> 
> ```
> # Allowed CORS origins
> #     If not specified, only the host runnig the server is allowed
> #     Allowed values are: ['*'], ['some-address.xyz', 'other-address.xyz'], []
> #     More information on the meaning of these values can be found at
> #     https://python-socketio.readthedocs.io/en/latest/api.html#asyncserver-class
> cors_allowed_origins = *
> ```

This is also not relevant. I understand why you included it, but this is the CORS settings of the PA server, it's the CORS settings of servers hosting images that were relevant for the original implementation not working. CORS is no longer relevant when the server does the request.

> EDIT:
> 
> Ahh now this is interesting. The pasted link turns into an `href`. The formatted link becomes an image `src`
> 
> Screenshot from 2024-07-01 11-45-38

Yeah it's just rendering as a url instead of an image, I'll investigate.

Kruptein:
Ah I see, I renamed the socket event at the end, and didn't rename it in all the proper locations 🤦

develroo:
> Ah I see, I renamed the socket event at the end, and didn't rename it in all the proper locations 🤦

It's a Monday..It happens ;)

Kruptein:
Should be fixed now

develroo:
Yup, works now. I'll adjust the documentation to reflect this and replace some with most. 


Cheers.

PR #1455, opened 2024-08-10
feat(dice): New dice features & changes

Kruptein:
This is a big PR that reworks a lot of the dice code.

## Technical changes

Notably, babylonjs (the 3d engine PA uses for the dice rolls) was upgraded from version 4 to 7.
This also changes the underlying physics engine from ammo to havok.

@planarally/dice has seen a huge restructure + rework of logic.

More changes to the dice repo are expected in the near future, 1 avenue I might be considering is to move all or most of the dice code inside of the core repo to the dice repo and have dice become a PA mod instead of being bundled with the main bundle.

## New UI

The dice tool UI has changed big time as a whole set of new clicky options is available to configure dice rolls with more advanced things you might be used to from other tools. (e.g. "keep highest 1").

Another noticeable new thing, is that now there is a switch to enable or disable the use of 3D dice.
By default no 3D will be used, and results will just be generated on the fly, allowing quick and light dice rolling.

The 3D assets and files will not load until you've flipped the 'use 3D' switch once.

The roll history has a slightly different UI, clicking on an entry will show the roll details again.
Similarly clicking on a notification that reports another player's roll, will do the same.


develroo:
Cool stuff.. But are we any closer to deciding on a d100 solution?  ;)

develroo:
Hmm  Just tested it. Looks good.. But aren't the settings supposed to stick between each activation ? Seem if I click to any tool and then back to dice, it resets to no 3D dice and Share with all. That could get annoying pretty quick if you roll without thinking.

Kruptein:
> Cool stuff.. But are we any closer to deciding on a d100 solution? ;)

I assume you mean rolling a d10 as well? that's on my todos as well indeed :)

> Hmm Just tested it. Looks good.. But aren't the settings supposed to stick between each activation ? Seem if I click to any tool and then back to dice, it resets to no 3D dice and Share with all. That could get annoying pretty quick if you roll without thinking.

Ah thanks for the feedback, that's not supposed to happen indeed

develroo:
> I assume you mean rolling a d10 as well? that's on my todos as well indeed :)

Ahh cool I see a d100 (10s) and so a d10 + d100 = % die 

Cheers

PR #1456, opened 2024-08-10
feat(Spell): Render hexes in spell tool when using hex grids

Kruptein:
When in hex grid mode, it makes more sense to have the spell tool output hexes instead of squares.

It's mostly still WIP for two reasons:
- FA does not have a free hexagon icon, I'm currently just using the square icon, but that might be confusing
- It basically duplicates code for hex rendering from somewhere else, but to make it a Polygon shape instead of direct canvas calls

The latter should be unified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,10 @@ tech changes will usually be stripped from release notes for the public
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


Kruptein:
Duplicate code has been resolved.

PR #1477, opened 2024-10-19
feat(Notes): Add Notes button to player's MenuBar

rexy712:
I noticed that as a player, you cannot access the NoteManager from the MenuBar like you can as the DM. The player has access to the NoteManager in other ways (eg right click menu of shape), so I feel like it would be nice to have a consistent way to open for them. This would also allow them access if there are only notes without shapes attached.

### DIFF
--- a/client/src/game/ui/menu/MenuBar.vue
+++ b/client/src/game/ui/menu/MenuBar.vue
@@ -93,10 +93,6 @@ const openClientSettings = (): void => uiSystem.showClientSettings(!uiState.raw.
         <div style="width: 12.5rem; overflow-y: auto; overflow-x: hidden">
             <!-- CHARACTERS -->
             <Characters />
-            <!-- NOTES -->
-            <button class="menu-accordion" @click="toggleNoteManager">
-                {{ t("common.notes") }}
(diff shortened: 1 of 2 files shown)


client/src/game/ui/menu/MenuBar.vue
```diff
@@ -93,6 +93,10 @@ const openClientSettings = (): void => uiSystem.showClientSettings(!uiState.raw.
         <div style="width: 12.5rem; overflow-y: auto; overflow-x: hidden">
             <!-- CHARACTERS -->
             <Characters />
+            <!-- NOTES -->
```
Kruptein, line 96: I would prefer to keep the location it has in the DM menu (i.e. below assets), as for me as a dm the location of the assets section, both because it's one of the most used ones and people might have some muscle memory where to click, but also because the list can expand down vertically more when it's higher.

I do agree that notes should be visible to players though, so could you end the `dm if clause` before notes and re-open it afterwards instead?

rexy712: Totally makes sense, will do

PR #1482, opened 2024-10-20
feat(Notes): Add ability to pop-out notes

### LINKED ISSUE #1426: [FEAT] Make pop out modal so they can be dragged outside of the browser window.
The title really says it all, because dragging notes onto another desktop is quite common so they are not cluttering the game screen.

I am pretty sure it is doable because of this [Stackoverflow post](https://stackoverflow.com/questions/65022204/make-a-popup-modal-dialog-movable-draggable-once-it-has-appeared) but unsure how much effort it woudl take ?

As it stands, dragging a note outside of the browser window does sorta pin it to the side but it is still no great. 

[Screencast from 2024-04-05.webm](https://github.com/Kruptein/PlanarAlly/assets/7777993/b540a581-f7f5-4e20-a87e-1d656f610981)

Anyway, just a thought.

Kruptein:
Closes #1426

This relies entirely on magic that I'm not sure is 100% intended to be used, so it might break at a certain point, but this is the only way to actually achieve it I believe.

Chances that some things will not work 100% are greater than 0, but basic functionality seems to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ tech changes will usually be stripped from release notes for the public
 
 ## Unreleased
 
+### Added
+
+-   Notes:
+    -   Notes can now be popped out to a separate window
+
 ### Changed
 
 -   Dashboard:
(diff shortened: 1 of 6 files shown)


develroo:
Can I ask how you tested this? I don't seem to get it to work. Do I still just use the pop out icon, or do I need to enable some sort of setting? 

FWIW I have tried both Firefox and a Chromium based browser.

PR #1483, opened 2024-10-23
Feature/audio stream

kyoaku:
This could be considered as a starting point for integrating sound into the project.

At the moment, it only searches for MP3 files in the root of the assets directory.

The DM will be able to preview an MP3 and start playing it (or another) for the players. They will also have the option to play the audio in a loop for the players.

image

On the players' side, they will be able to pause, resume playback, and adjust the volume.

image


Kruptein:
Hey, thanks for taking the time to submit this PR, a quick glance to the changes seems to indicate that you found your way around the codebase which is always a happy surprise 😅 

That said, I haven't reviewed the changes yet, because I'm a bit conflicted on whether this is a good fit for the main codebase or better suited as a mod[^1].

This is often a narrow line and not always easy to judge for me. Audio is a feature that can be prominent in ttrpgs, but I'm not sure how many people actually use mp3s as such compared to relying on other services like youtube/spotify or more ttrpg focused things like tabletopaudio and the likes.

So I do think some common audio apis should exist in the core codebase to facilitate an unambigous api to integrations with other services (e.g. triggering certain audio effects when an area is entered etc, stopping/resuming an audio stream, changing audio scenes). 

As a mod/plugin (ideally) you would have all the power to tweak it to your own liking, whereas in the main codebase you have to convince me :D

If this were to come to the main codebase my biggest input would be related to the DM side of things. I know that it's a starting point, but I would not put the configuration in the menubar but rather do it like notes with a big modal somewhat, allowing you to see a list of audio files and being able to search through them instead of having 1 big select dropdown which is quickly going to be confusing if you have a lot of files. This can in the future also act as a tabbed interface where you can tab between potential audio sources (e.g. uploaded / spotify / ...).

I'm also not sure what the benefits are of showing the audio player to players. Being able to control the volume is something that should be possible, but pausing/resuming audio seem less relevant to me.
What happens if a player paused their audio, is it paused for everybody? If not what if they later resume the playback, but the DM changes to another audio file.

[^1]: The plugin/mod API is still in early stages, which doesn't make it any easier. See https://github.com/Kruptein/PlanarAlly/pull/1297 for more info.

Issue #1022, opened 2022-08-10
[FEAT] Remove file extension from uploaded assets


FuzzySadist:
Remove file extension from uploaded assets.  Some of the file names are long enough without adding the 4 extra characters.  As with my previous request.... I nest folders and upload descriptive names, so the .jpg etc just makes it that much harder to deal with.

PR #1491, opened 2024-10-29
feat(AssetManager): Allow editing asset names without a popup

rexy712:
This allows you to click on the name of an asset or folder and simply modify the name in place, rather than having a popup appear to change the name. You can complete editing by simply pressing Enter or clicking outside of the editable text.

The popup system works fine, but I find it to be cumbersome and less intuitive. It also has the issue of starting with a blank text field instead of the current name of the asset/folder. Most of the time when I'm editing a name, it's to fix a small typo I made, so having to start at a blank slate each time is frustrating.

While this may not be a perfect implementation, it's more similar to what most people would expect after using a modern file browser. I also do not remove the Rename functionality in the context menu in this change so that not everyone is forced into using this system if they're used to the old way.

https://github.com/user-attachments/assets/a7c0eb46-725f-4ca7-bb0c-1cce35db030b


Kruptein:
It's interesting because to me this is less intuitive :D the effect is very subtle which makes it easy to accidentally click on a name when you just want to select something.

I'm also very curious as to your statement
> It's more similar to what most people would expect after using a modern file browser

I personally haven't seen this being used anywhere as a default :p Neither windows explorer or the file managers I'm used to on linux have this behaviour by default, nor do online file storage solutions like a google drive. I'm aware they can usually be configured to do so however, but I wouldn't assume this to be done by most people.

That said, I do agree that the context menu and action popups can use some UI/UX improvements.

I think the bigger issue however is that the extensions are still present on the asset names (https://github.com/Kruptein/PlanarAlly/issues/1022), which, I'm going to guess, is the biggest reason why you might need to rename them so often. If that were fixed, would you still have a need to rename assets that often causing the context menu to remain a bother?

rexy712:
> It's interesting because to me this is less intuitive :D the effect is very subtle which makes it easy to accidentally click on a name when you just want to select something.

Hm I hadn't thought it would be easy to accidentally click. I always aim for the image when trying to select, but fair point.

> I personally haven't seen this being used anywhere as a default :p Neither windows explorer or the file managers I'm used to on linux have this behaviour by default, nor do online file storage solutions like a google drive.

I suppose I would say most file managers I've used allow a slow doubleclick to edit the name. I went for what seemed easiest to me, which is similar but not the same as a file manager.

> I think the bigger issue however is that the extensions are still present on the asset names (#1022), which, I'm going to guess, is the biggest reason why you might need to rename them so often. If that were fixed, would you still have a need to rename assets that often causing the context menu to remain a bother?

That would take care of a good amount of the renaming certainly.

Perhaps a better implementation would be a combination of the old method and this. Instead of clicking the title to initiate a rename, use the context menu option to make the title editable in place. Would that be worth trying?

Kruptein:
> Perhaps a better implementation would be a combination of the old method and this. Instead of clicking the title to initiate a rename, use the context menu option to make the title editable in place. Would that be worth trying?

That would definitely be an upgrade over what it currently does yeah,
so I think that's a good compromise for now.

rexy712:
I changed the way this works to now doing a right-click->rename to make the title editable in-place. The newly editable title gains focus and the text is all selected by default.

One issue I have is that there is a linter error I cannot seem to figure out how to fix. It has to do with a part of the file I didn't change. The `remove` function in client/src/assetManager/AssetContext.vue was completely untouched, but now its inclusion in the `sections` array causes a `Promise-returning function provided to property where a void return was expected @typescript-eslint/no-misused-promises` linting error. I checked the definition of the types in that array, and it should still be fine, as the functions are allowed to return a void or Promise<void>.


https://github.com/user-attachments/assets/21fba7c7-2f1b-4964-b7bc-fca9c5fcea83

Kruptein:
So what is happening with the lint error:
There are 3 sections, they all have roughly the same shape, except 'share' which also has a disabled property. Previously it so happened that the 2 sections that did not specify `disabled` all were `async` and the one that did specify `disabled` was sync, so typescript autogenerated a distinct set of types were the presence of `disabled` dictates whether the `action` is sync or async.

Now this is no longer the case and apparently it's not happy. I'm not entirely sure why it's not just making the `action` type a union, but that's what's happening.

The solution is rather easy, explicitly tell the computed what type it is:
```typescript
const sections = computed<Section[]>(() => [
   ...
]);
```
You can import `Section` from the core types (see also `ContextMenu` props)

client/src/dashboard/Assets.vue
```diff
@@ -31,6 +31,9 @@ const body = document.getElementsByTagName("body")[0];
 const activeSelectionUrl = `url(${baseAdjust("/static/img/assetmanager/active_selection.png")})`;
 const emptySelectionUrl = `url(${baseAdjust("/static/img/assetmanager/empty_selection.png")})`;
 
+const currentRenameElement = ref<HTMLDivElement | null>(null);
+const titleRefs = ref<Record<AssetId, HTMLDivElement | null>>({});
```
Kruptein, line 35: I would make this either a `reactive<Record<AssetId, HTMLDivElement>>` or a `ref<Map<AssetId, HTMLDivElement>>`.

Given that it's an object, using `reactive` will be just a tad bit more ergonomic as you don't have to do the `.value` in the `setTitleRef` function. On the other hand I also just like a `Map` when it's an actual mapping over an object in which case a `ref` is needed.

Feel free to choose what you prefer

client/src/dashboard/Assets.vue
```diff
@@ -265,6 +288,49 @@ function canEdit(data: AssetId | DeepReadonly<ApiAsset> | undefined, includeRoot
...
+    const range = document.createRange();
+    range.selectNodeContents(el);
+    const sel = window.getSelection();
+    if(sel) {
+        sel.removeAllRanges();
+        sel.addRange(range);
+    }
+}
+
+function handleRenameEvent(id: AssetId) : void {
```
Kruptein, line 321: I would rename this maybe to something like `showRenameUi`.

client/src/dashboard/Assets.vue
```diff
@@ -265,6 +288,49 @@ function canEdit(data: AssetId | DeepReadonly<ApiAsset> | undefined, includeRoot
...
+    const el = titleRefs.value[id];
+    if (el) {
+        currentRenameElement.value = el;
+        nextTick().then(() => {
```
Kruptein, line 325: This is fine, but you can also make the function async and do regular `await`.

client/src/dashboard/Assets.vue
```diff
@@ -265,6 +288,49 @@ function canEdit(data: AssetId | DeepReadonly<ApiAsset> | undefined, includeRoot
...
+    const el = titleRefs.value[id];
+    if (!el) {
+        return false;
+    }
+
+    return el === currentRenameElement.value;
+}
+
+function selectElementContents(el: HTMLElement) : void {
+    const range = document.createRange();
```
Kruptein, line 312: I had no idea about this API, cool!

client/src/dashboard/Assets.vue
```diff
@@ -265,6 +288,49 @@ function canEdit(data: AssetId | DeepReadonly<ApiAsset> | undefined, includeRoot
...
 }
+
+function setTitleRef(el: HTMLDivElement | null, id: AssetId) : void {
+    if (el) {
+        titleRefs.value[id] = el;
+    } else {
+        delete titleRefs.value[id];
+    }
+}
+function canRename(id: AssetId) : boolean {
```
Kruptein, line 299: I think we can drop this function.

Change `currentRenameElement` to be `currentRenameAsset` instead (and thus track `AssetId` instead of `HTMLDivElement`) and we can just do `:contenteditable='currentRenameAsset === folder.id`.

The `canEdit` check will already be done when the `renameAsset` function is called and the context menu open is also guarded by the `canEdit` function.

This would also mean we actually don't need to track the `titleRefs` at all.

Having `contenteditable` be a direct reactive expression instead of a function is going to be more performant as it won't have to rerun the function on every render. So where possible that's always a benefit.

Kruptein (changes requested):
Works well! I have a bunch of comments but they're almost all purely nitpicking about small stylistic thing.

I recommend to first check my comment on the `canRename` function as some of my other comments may no longer be relevant if you agree with my feedback there.

client/src/dashboard/Assets.vue
rexy712: True I could change that to currentRenameAsset, that would make sense. Getting rid of the function is also much nicer, I agree.

> This would also mean we actually don't need to track the titleRefs at all.

So the reason I added in `titleRefs` was because I needed a way to get the div element so that I could call `focus()` on it and select its contents. I didn't see a way to accomplish this without this.

client/src/dashboard/Assets.vue
rexy712: I went with reactive. It was nicer to not have to say `.value` everywhere.

rexy712:
> I'm not entirely sure why it's not just making the `action` type a union, but that's what's happening.
> 
> The solution is rather easy, explicitly tell the computed what type it is:
> 
> ```ts
> const sections = computed<Section[]>(() => [
>    ...
> ]);
> ```
Thanks for the solution! It was driving me crazy. This gives me flashbacks to C++ template errors and trying to debug those, but this time in a much less familiar language.

client/src/dashboard/Assets.vue
```diff
@@ -168,6 +171,26 @@ function openContextMenu(event: MouseEvent, key: AssetId): void {
...
+    if(target.textContent === null){
+        target.textContent = oldName;
+        currentRenameAsset.value = null;
+        return;
+    }
+    const name = target.textContent.trim();
+
+    if (name === undefined || name === "" || name === "..") {
+        target.textContent = oldName;
+    }else{
```
Kruptein, line 189: I'm surprised the linter doesn't pick this up, but add some spacing around the else.

Maybe also make it an `else if (name !== oldName)`. The extra network call isn't going to be a big issue, but nevertheless we know it's not necessary.

client/src/dashboard/Assets.vue
Kruptein: Ah I see, I missed that use.

It feels a little bit overkill to store a ref to every single asset when the chances are rather slim that we actually ever are going to interact with the majority of them.

We could store the currently selected asset (either by click or rightclick) in a variable I suppose. As we can get the element from the click/contextmenu events.

That said, it's not the end of the world either and usually you're not super long in the asset manager anyway, so we probably can get away with it.

I leave it up to you whether you want to change this or keep it, I'm fine with either decision.

Kruptein (changes requested):
Last miniscule change :D

client/src/dashboard/Assets.vue
rexy712: I was having issues because openContextMenu was being triggered by an event generated by the surrounding div, not the title's element itself. I found a way to retrieve the child element though, so I'd also like to get rid of `titleRefs` entirely.

client/src/dashboard/Assets.vue
rexy712: Makes sense, fixed!

PR #1500, opened 2024-11-05
feat(NoteManager): Add ability to show all available notes

rexy712:
Currently the NoteManager allows one to filter by local notes or global notes. My problem is that sometimes I just want to see all available notes, regardless of if they're global or local.

I was running a session today where I was swapping between notes during an event, and had trouble tracking some down because I was searching in local when I should have searched in global, or vice versa. I believe it would have been easiest to see all notes and filter them down to what was relevant with the search function.

This pull request adds the "all" option for note filtering and replaces the `ToggleGroup` with a `select` element, as it seems less cluttered than having 3 buttons.

Current dev branch:
Screenshot_20241105_145334

This pull request:
Screenshot_20241105_145403
Screenshot_20241105_145441


Kruptein:
Hm, I'm not entirely sure yet what I think of this. Visually it feels like a downgrade (which can undoubtedly be improved) and I understand the functional benefit, but it seems like in practice you're just always going to be on the `all` option if I make some assumptions, so at that point what actually is the added value of the dropdown.

The main reason I'm a bit reluctant is because I'm honestly still a bit dissatisfied with the current local/global split in general. In my original vision the vast majority of notes would be local (i.e. campaign specific) and I thought of global notes as something mostly reserved for things like monster stats that, once attached to a monster template, would not really be something you end up looking for with the note manager often and instead access through the shape itself.

You seem to hint at using a bunch of global notes yourself, so it would be interesting to me if you could maybe share what kind of notes you have in the global section, that are frequently accessed during a game that are not campaign specific.

_Small side tangent incoming_

In the meantime I've dabbled with a variety of systems and I've also become more aware that global is actually a bit too wide. I have no need to see all my d&d related global notes when I'm in a mothership campaign, nor do I need to see the rules for lancer.  I could "solve" this by giving tags to each of my notes that link them to a certain system for example, but I would have to filter on that tag manually every time.

So I think it might become time to add some new setting to campaigns that allows you to define a "system". That can be used to reduce the amount of notes in the global notes section, but also for other things like configuring dice sets, ruler settings etc.

_(A similar problem actually also exists in the assets space. Currently the entire asset listing is present and people can organize them however they want, which I think is nice, but at the same time, I don't need most assets that are related to other campaigns or systems.)_

_end of side tangent 1, start of side tangent 2_

There is also another problem with global notes that I found out recently related to default access permissions. Given that they are campaign agnostic, `default view access` truly grants view access to _all_ users on the server to that note. Which for multiple reasons is not great.

One option is to retrofit 'default view access' to mean 'every user that shares a campaign with the note author can view the note'. Which is straight up better than what it currently does, but it still feels a bit too "abuseable".

Another option is to just remove the default settings on global notes. It would currently just be a bit of a pain to manually add multiple users to a specific note, let alone multiple. So if I go this route (which seems the most likely), I'll probably need to add some "add all players in the campaign" button to the note access settings

_end of side tangent 2_

Sorry for the big ramble as most of it is not really related to this PR specifically. But before tackling note UI changes, I kinda want to have a better idea of which direction notes are headed in general. I like what notes currently offer, but it still feels off in some regards.

rexy712:
> Visually it feels like a downgrade (which can undoubtedly be improved)

I was also looking this over again yesterday and thinking that it needed to be properly styled to fit the rest of the theme. I don't have much skill at visual design, but could try improving this if it seems worth the effort.

> it seems like in practice you're just always going to be on the `all` option if I make some assumptions

Probably a fair point. I don't really distinguish between the note type once they're made and I just want to find them. While making them, I like to put them in the proper category so I can reuse statblocks and such in future campaigns.

> so at that point what actually is the added value of the dropdown.

I didn't want to take away the ability to filter by local/global if someone else wanted to use them, thus the dropdown. 

<br>

> I thought of global notes as something mostly reserved for things like monster stats that, once attached to a monster template, would not really be something you end up looking for with the note manager often and instead access through the shape itself.

I would say this is how I access monster stats most of the time. The problem being that not all global notes are necessarily tied to a shape.

> You seem to hint at using a bunch of global notes yourself, so it would be interesting to me if you could maybe share what kind of notes you have in the global section, that are frequently accessed during a game that are not campaign specific.

Most of my global notes are also statblocks, though not always associated with a shape. For example statblocks for hazards, traps, etc. I don't usually draw shapes for these kind of things, I'll just have the note open to review. There are also a few global notes that describe game subsystems, for example the pathfinder influence subsystem. I also keep these open for quick review and share them with players so they can do the same.

<br>

> So I think it might become time to add some new setting to campaigns that allows you to define a "system". That can be used to reduce the amount of notes in the global notes section, but also for other things like configuring dice sets, ruler settings etc.

Hm, that's an interesting idea. That would definitely make sense to me. The note would have to have a way to select what system it's a part of. If a system is not chosen for it, it shows up for all.

While on the topic, I've had a couple players note that the 'cone' shape in the spell tool is not the correct angle for pathfinder. I'm sure a bunch of games define what a 'cone' is differently, so that could also be a setting within a 'system'.

> _(A similar problem actually also exists in the assets space. Currently the entire asset listing is present and people can organize them however they want, which I think is nice, but at the same time, I don't need most assets that are related to other campaigns or systems.)_

My current organization system for this is having a folder for generic things and folders for campaign specific things. While I understand that many of the assets in campaign specific folders won't be used in any other setting, I still don't like the idea of them being hidden so that I can't see them. There have been times where I dig out an old NPC from some other old game we played and have them make an appearance as a joke.

<br>
 
> There is also another problem with global notes that I found out recently related to default access permissions. Given that they are campaign agnostic, `default view access` truly grants view access to _all_ users on the server to that note. Which for multiple reasons is not great.

I hadn't thought of this since I just run the server for me and my friends who all share the same campaigns, but definitely not the ideal setup in that case. The feeling the 'default' button gave me was sharing with all users in the current campaign, not 'any user that shares a campaign with me'. So I think it would be better to remove the default option in this case. The 'add all players in this campaign' button would be good too.

<br>

> Sorry for the big ramble as most of it is not really related to this PR specifically. But before tackling note UI changes, I kinda want to have a better idea of which direction notes are headed in general. I like what notes currently offer, but it still feels off in some regards.

No, it's good to know. I was working on some ideas for NoteList to hopefully allow easier tag searching, but I can put that off if you want to figure out the overall idea of note implementation.

Kruptein:
> > Visually it feels like a downgrade (which can undoubtedly be improved)
> 
> I was also looking this over again yesterday and thinking that it needed to be properly styled to fit the rest of the theme. I don't have much skill at visual design, but could try improving this if it seems worth the effort.

If we go ahead with this, I can give it a shot as well. I'm not claiming to be great at design either (PA's styles are all over the place), but I can usually make something work :V

> > it seems like in practice you're just always going to be on the `all` option if I make some assumptions
> 
> Probably a fair point. I don't really distinguish between the note type once they're made and I just want to find them. While making them, I like to put them in the proper category so I can reuse statblocks and such in future campaigns.
> 
> > so at that point what actually is the added value of the dropdown.
> 
> I didn't want to take away the ability to filter by local/global if someone else wanted to use them, thus the dropdown.

That's fair

> > I thought of global notes as something mostly reserved for things like monster stats that, once attached to a monster template, would not really be something you end up looking for with the note manager often and instead access through the shape itself.
> 
> I would say this is how I access monster stats most of the time. The problem being that not all global notes are necessarily tied to a shape.
> 
> > You seem to hint at using a bunch of global notes yourself, so it would be interesting to me if you could maybe share what kind of notes you have in the global section, that are frequently accessed during a game that are not campaign specific.
> 
> Most of my global notes are also statblocks, though not always associated with a shape. For example statblocks for hazards, traps, etc. I don't usually draw shapes for these kind of things, I'll just have the note open to review. There are also a few global notes that describe game subsystems, for example the pathfinder influence subsystem. I also keep these open for quick review and share them with players so they can do the same.

Thanks for the input!
 
> > So I think it might become time to add some new setting to campaigns that allows you to define a "system". That can be used to reduce the amount of notes in the global notes section, but also for other things like configuring dice sets, ruler settings etc.
> 
> Hm, that's an interesting idea. That would definitely make sense to me. The note would have to have a way to select what system it's a part of. If a system is not chosen for it, it shows up for all.
> 
> While on the topic, I've had a couple players note that the 'cone' shape in the spell tool is not the correct angle for pathfinder. I'm sure a bunch of games define what a 'cone' is differently, so that could also be a setting within a 'system'.

Yeah it's the same deal with some requests for the ruler to work with alternate rules depending on systems/settings. I can add all these extra settings regardless of a system concept, but it could help with not overwhelming new PA users to see a million different ruler settings when only 2 are potentially relevant for their system.

> > _(A similar problem actually also exists in the assets space. Currently the entire asset listing is present and people can organize them however they want, which I think is nice, but at the same time, I don't need most assets that are related to other campaigns or systems.)_
> 
> My current organization system for this is having a folder for generic things and folders for campaign specific things. While I understand that many of the assets in campaign specific folders won't be used in any other setting, I still don't like the idea of them being hidden so that I can't see them. There have been times where I dig out an old NPC from some other old game we played and have them make an appearance as a joke.

Good point, I have some design in my head where you could select some folder as your "main campaign folder" which would be the default view, but still be able to navigate to other folders and access all your assets. (This would tie in with a move to have access to assets in-game through a bigger modal like notes instead of the current sidebar)

> > There is also another problem with global notes that I found out recently related to default access permissions. Given that they are campaign agnostic, `default view access` truly grants view access to _all_ users on the server to that note. Which for multiple reasons is not great.
> 
> I hadn't thought of this since I just run the server for me and my friends who all share the same campaigns, but definitely not the ideal setup in that case. The feeling the 'default' button gave me was sharing with all users in the current campaign, not 'any user that shares a campaign with me'. So I think it would be better to remove the default option in this case. The 'add all players in this campaign' button would be good too.

Specifically regarding the line "The feeling the 'default' button gave me was sharing with all users in the current campaign, not 'any user that shares a campaign with me'". For global notes there is no related campaign really. Sure you made the note in campaign X, but because you made it global, you can also access it from campaign Y. So if those 2 campaigns have a complete different set of players, who does "default view" apply to?

I'm obviously not saying that your intuition was "wrong". The term `default` is simply ambiguous in the context of a global note and I just wanted to highlight way it works the way it works currently. But we both seem to agree on removing the default section for global notes.

It does however also highlight that the concept of "global" is not super clear I guess.

> > Sorry for the big ramble as most of it is not really related to this PR specifically. But before tackling note UI changes, I kinda want to have a better idea of which direction notes are headed in general. I like what notes currently offer, but it still feels off in some regards.
> 
> No, it's good to know. I was working on some ideas for NoteList to hopefully allow easier tag searching, but I can put that off if you want to figure out the overall idea of note implementation.

It's certainly welcome, I'm not intending to completely change how the note system works. There is just something to it currently that is lacking and I can't quite put into words what it is exactly :D

Improving the tagging can definitely work as both adding tags is a bit lacking currently (e.g. no suggestion of existing tags etc), and searching for tags is possible, but it also searches in other places, so yeah I think some improvements with some special search syntax and/or just allowing you to click on a tag would help a lot. But I'll see what you come up with :p

You're always free to make a draft PR with some less fleshed out ideas/concepts if it's something you're not sure about or don't want to invest time in if you're not sure if I'll like it :D


I guess we can move ahead with this PR, with either you or me adding some style changes. (I still have to check the actual code changes as well).

Another small side tangent is that I might have to swap to server side pagination + searches instead of just loading everything on the client and filtering/searching there.

rexy712:
I took a shot at making the styling better. I tried to keep with the same vibe the toggle switch gave off; same color, same rounded, black border. Added some padding to make it feel a bit more important.

image

Kruptein:
That looks better already! I'm only not sure yet whether I prefer All or Local as the default option.

rexy712:
The reason I have it currently set to All rather than Local is because I thought it would be good to default to no filters and add them if desired. But I can see the argument for making Local default, since those are more likely to be relevant. I'm good with either option.

Kruptein:
I think defaulting to 'local' makes more sense to me, but we can add a localStorage setting that remembers the last choice so that people who prefer 'all' don't have to change it every session.

PR #1508, opened 2024-11-10
feat(NoteManager): add a clear button to the searchbar

rexy712:
This adds an 'X' button on the right side of the searchbar in the NoteManager which will clear the current search query. It's never been a huge issue to not have it, but it's a pretty common feature for search boxes that I've gotten used to. This is just a minor change that I would like to see.

### DIFF
--- a/client/src/fa.ts
+++ b/client/src/fa.ts
@@ -25,6 +25,7 @@ import {
     faChevronUp,
     faCircle,
     faCircleInfo,
+    faCircleXmark,
     faClockRotateLeft,
     faCog,
     faCogs,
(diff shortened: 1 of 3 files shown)


client/src/fa.ts
```diff
@@ -153,6 +154,7 @@ export function loadFontAwesome(): void {
         faVideo,
         faWindowClose,
         faWindowRestore,
+        faX,
```
Kruptein, line 157: I think circle-xmark would be better.

rexy712: After trying it out, I agree

Issue #1495, opened 2024-10-31
[BUG] Other window overlapping Note Manager window render order incorrect


rexy712:
**Describe the bug**
When both the Note Manager window and another window are open at the same time, if the other window is in the foreground and overlaps the Note Manager's search field, the search field is rendered on top of the other window. The rest of the Note Manager is rendered behind.

**To Reproduce**
1. Open a window, a Note or Edit Shape window, I'll say Edit Shape for this demonstration.
2. Open the Note Manager
3. Make the Edit Shape window the foreground window
4. Drag the Edit Shape window over the Note Manager search bar.
5. Search bar is drawn over the Edit Shape window.

**Expected behavior**
The whole Note Manager should be drawn behind the other window.

**Screenshots**
![image](https://github.com/user-attachments/assets/f885d997-b80b-45b7-a3ca-c1d5e7ccbe62)

**Desktop:**
 - OS: gentoo linux
 - Browser: ungoogled-chromium 129.0.6668.89
 - Version: dev branch commit [9562e0c](https://github.com/Kruptein/PlanarAlly/commit/9562e0c3fac89629ad52b7013d7af4c115f54611)

PR #1509, opened 2024-11-10
bug(NoteManager): Fix search options dialog not overlapping some elements

rexy712:
In #1505 the z-index property was removed from the div inside #note-search, which corrected the issue of the searchbar rendering overtop other modals. However, it introduced the problem of having the headers of the note listing render overtop the search options dialog.

| Dev Branch | This PR |
| ----------------  | ---------- |
| Old | New

I addressed this issue by reintroducing the z-index, but only to the settings dialog. This of course reintroduced part of the original problem from #1495, as well as making the dialog render overtop its own toggle control button. So I add a copy of the toggle to the dialog and make the original toggle button no longer `position: absolute` since there's no point to doing so anymore.

This left the issue of the dialog still drawing over other modals when they were moved to its position. I decided to solve this in a way that would solve another of my gripes. I call this a dialog window because that's how it feels to me; it should only be open for a short while and close as soon as it loses focus. By having the dialog close when it loses "focus", it will no longer draw over anything it shouldn't because it's already closed! This feels much better to me in general and it solves the render order drawing issue.

This completes solving #1495.


client/src/game/ui/notes/NoteList.vue
```diff
@@ -45,6 +46,22 @@ const shapeName = computed(() => {
...
+function handleClickOutsideDialog(event: MouseEvent): void {
+    if (searchOptionsDialog.value) {
+        if (showSearchFilters.value && !searchOptionsDialog.value.contains(event.target as Node)) {
+            showSearchFilters.value = false;
+        }
+    }
+}
+
+onMounted(() => {
+    document.addEventListener('mousedown', handleClickOutsideDialog);
```
Kruptein, line 58: I know it's not present everywhere in PA, but I would change this to be `pointerdown`. It's a small change, but then it will work on both touch and mouse devices in general.

rexy712: Ah makes sense. Will do

Kruptein:
Small merge conflict since merging the clear button

PR #1510, opened 2024-11-10
feat(NoteManager): Improve filtering by tags in note search

rexy712:
This is one of the ideas I was talking about in https://github.com/Kruptein/PlanarAlly/pull/1500#issuecomment-2466251120.

I tried my hand at revamping the NoteManager's tag searching capabilities. I went through several ideas before deciding on this solution. I tried integrating tag bubbles in the search bar, tried implementing an autocomplete feature to the main search bar, and several other ideas that seemed good in concept but fell flat in reality.

This PR makes a few changes. Firstly the new functionality: Tag filtering is separate from the search bar entirely. I believe that this makes the search bar feel more accurate because it will by default only search for titles, which is what I want most of the time. Tags are instead handled in the "Filter" section just below the search bar.

The new "Filter" section contains bubbles with the tag names which can be clicked to remove. It also has a search bar on the right side (hidden behind a '+' icon initially to keep a minimal look) to search through existing tags and add/remove them by clicking on their respective bubbles in a dropdown menu. They can also be iterated through using the arrow keys (see video 1).

I'm not sure if the '+'/'-' button is really necessary, but it seems to be a cleaner look to me, so I left it in. It would be simple enough to remove and just have the search bar always visible if so desired.

I also believe that the tag searchbar can be reused in the NoteEdit window to add tags to a shape. Though it would need to be able to add new tags as well as search for ones that already exist. The general idea would be pretty similar though. That implementation lies outside the scope of this PR though.

Another way to toggle tags in the new "Filter" section is with the tag bubbles in the note list. These bubbles are also now clickable, toggling the tag's status as a filter in the search (see video 2).

I'm not sure they should be able to toggle the tag or should just be used to enable the tag. I left it as a toggle until I got some feedback about this idea in general.

Finally, the new "Filter" section also contains the name of the selected shape, if any, instead of putting in the search bar. It is always the leftmost object in the "Filter" area. This seems to make sense to me and should make it more clear that the name can be interacted with to remove it as a filter. When the NoteManager first came about, I didn't know that clicking on the shape name next to the search bar would remove the shape filter. I've had players also say they didn't know until I told them (see video 3).

https://github.com/user-attachments/assets/6c6ac5b0-c06b-4e8f-9f1a-1251db21e15c

https://github.com/user-attachments/assets/dbf8c5da-c80c-411d-9a0d-792a039b7bca

https://github.com/user-attachments/assets/6529991b-c6dd-4691-a75a-ee1732b70309

### DIFF
--- a/client/src/game/ui/notes/TagAutoCompleteSearch.vue
+++ b/client/src/game/ui/notes/TagAutoCompleteSearch.vue
@@ -201,12 +201,12 @@ function handleEnter(): void {
     flex-wrap: wrap;
     row-gap: 0.5rem;
     padding: 0.5rem;
-    margin: 0;
+    margin: 0 0.7rem;
     border: 1px solid black;
     border-top: 0;
     box-shadow: 0 0.1rem 0.875rem 0 #cacaca;
     background: white;
-    width: 100%;
+    width: calc(100% - 1.4rem);
     z-index: 1;
 }
 
--- a/client/src/game/ui/notes/TagAutoCompleteSearch.vue
+++ b/client/src/game/ui/notes/TagAutoCompleteSearch.vue
@@ -195,16 +195,21 @@ function handleEnter(): void {
 
 
 #autocomplete-results {
-    position:absolute;
+    position: absolute;
     display: flex;
+    padding: 0.5rem;
+    margin: 0 0.7rem;
+
     flex-direction: row;
     flex-wrap: wrap;
     row-gap: 0.5rem;
-    padding: 0.5rem;
-    margin: 0 0.7rem;
+
     border: 1px solid black;
     border-top: 0;
-    box-shadow: 0 0.1rem 0.875rem 0 #cacaca;
+    box-shadow: 0 0.1rem 0.675rem 0 rgba(0, 0, 0, 0.5);
+    border-radius: 0 0 0.5rem 0.5rem;
+    clip-path: inset(0 -1rem -1rem -1rem);
+
     background: white;
     width: calc(100% - 1.4rem);
     z-index: 1;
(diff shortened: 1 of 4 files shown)


develroo:
Looks very slick and reminds me of Nextcloud. 

Good job! I appreciate your efforts.

rexy712:
> Looks very slick and reminds me of Nextcloud.

Thanks! I never expected to be compared to a polished product like that

> FWIW the standard paradigm I have seen for tags is that when you hover over them, they have a small 'x' in the corner, with which you can remove the tag

Similar to how tag bubbles are currently implemented in the Note Edit screen? That's certainly possible to do as well.

> but if you click on the tab it can be disabled or enabled.

Not sure what you mean by this though.

develroo:
> > but if you click on the tab it can be disabled or enabled.
> 
> Not sure what you mean by this though.

Meant to say tag, but essentially you click on it once it goes darker, to signify disabled, click again it gets brighter to show it is enabled. Otherwise, I see the red x is there so that makes sense.

client/src/game/ui/notes/TagAutoCompleteSearch.vue
```diff
@@ -0,0 +1,227 @@
...
+
+const searchBar = ref<HTMLInputElement | null>(null);
+
+watch(() => props.modelValue, (newValue) => {
+    if (newValue !== undefined) {
+        searchQuery.value = newValue;
+    }
+});
+
+// https://stackoverflow.com/a/69901745
```
Kruptein, line 27: Is this comment still relevant? I don't really see a `as Ref<` anywhere?

Kruptein:
Works nice in general, my main complaint is that the autocomplete result div looks very disconnected

image

client/src/game/ui/notes/TagAutoCompleteSearch.vue
rexy712: I forget why exactly I added that comment now. Your'e right it's no longer relevant, I just forgot to remove it when I changed results to `computed`

rexy712:
Pulled in the corners of the results so they don't extend to the rounded ends of the search box.

image

Kruptein:
> Pulled in the corners of the results so they don't extend to the rounded ends of the search box.

That looks better indeed

PR #1518, opened 2024-11-23
i18n(UI): zh(95%)

SuikaXhq:
Replace 95% of the hardcoded texts with i18n functions. Translated into Chinese (Simplified).

DiceTool is excepted since there may be an upcoming update from @rexy712 (saw in Discord).


Kruptein:
The linting error is an unused export, which I'm going to address in a small update after merging this.

Issue #1061, opened 2022-10-01
[BUG]  3D rendered dice sit on top of menu dialogs


10leej:
**Describe 
![graphic-bug](https://user-images.githubusercontent.com/2258333/193422308-11100360-0225-4214-b4dc-f00f1d8d1f72.png)
![graphic-bug-2](https://user-images.githubusercontent.com/2258333/193422352-3b42b068-27e2-4e0a-8941-7a4acc915295.png)

**To Reproduce**
1. Roll a dice
2. open a window
> Works in reverse order too

**Expected behavior**
Windows should sit above dice as really the 3d render is sort've irrelevant after it's done animating

**Desktop (please complete the following information):**
 - OS: Void Linux
 - Browser Firefox 105
 - Version 2022.2.3

develroo:
Not sure this is a bug per se. I mean, I can easily reproduce it, but it has never really been an issue, as you can close the dice dialog box and they all vanish.. leaving the roll in the log. 

What behaviour are you expecting exactly? That they should somehow roll under open windows? Then you would not be able to see the result at all? So what is the point of having animated dice in that case anyway ?

Just curious.

10leej:
I think they should roll under. Really I care for the result not the flair. I also have a properties window open often too.

________________________________

develroo:
@10leej  Given that there is now a dice history, do you think it is fair to close this issue as obsolete now ?

10leej:
> @10leej Given that there is now a dice history, do you think it is fair to close this issue as obsolete now ?
The issue still persist so I'm of the belief that this should still stay open.

10leej:
![Image](https://github.com/user-attachments/assets/708c19fa-b43a-4844-bd27-3de689cc2677)

Issue still persists, thankfully there's a front facing option to disable the 3d render

Kruptein:
It's honestly weird to me to have them roll under.

It just means that some dice will not be readable at all because they're hidden behind some menus or UIs. At that point just roll in the smaller 3D box or use 2D.

Having a manual clear button or changing the timeouts to make them disappear faster is another matter that I can understand however.

develroo:
> ![Image](https://github.com/user-attachments/assets/708c19fa-b43a-4844-bd27-3de689cc2677)
> 
> Issue still persists, thankfully there's a front facing option to disable the 3d render

[...]

PR #1519, opened 2024-11-24
style(UI): Reskin the dice tool

rexy712:
This is an overhaul of the visuals of the dice tool.

### Last Results Section
The top of the tool window is dedicated to showing the previous roll, being likely the most important thing one would want to see. The result is displayed clearly on the lefthand side in a prominent black-bordered box and the breakdown is to its right. Each Part of the result is in a separate block with operators floating between. Each Part's block shows a full breakdown of what the rolled values were.

### History
The next part down is the history drawer. It is closed by default and can be opened by hitting the 'Full History +' toggle. The drawer starts small and grows up to a maximum size as it becomes populated by entries. Each entry represents a dice roll that the player did or was shared with them by another. An entry shows who performed it, what dice were rolled, and the grand total. Clicking on an entry opens a breakdown view with more details about what each dice's result was just like the breakdown view for the last results section. Just above the breakdown is a Reroll button. Clicking this will populate the input field with the input value that generated this roll to allow a player to quickly generate a result from the same roll string.

### Options Cluster
The options cluster is pretty straight forward. On the left is the control for using 3D dice. No and Yes are identical to the way the old dice tool worked. 'Box' will add a dice box to the top of the tool window. 3D dice will be thrown inside of this box instead of on top of the gameboard when this option is enabled. This prevents issues like #1061 and is more aesthetically pleasing in my opinion. 

On the right of the options cluster is the share toggle, which is identical to the old dice tool. Nothing to add here.

### Advanced Options
The advanced options are the same as in the old dice tool layout, but they are rendered more as inside a drawer to visually distinguish what is a part of the 'advanced' menu. These options are also no longer between the literal digits and the rest of the input buttons, which felt a bit weird.

### Literal and Dice Selection
The only change here is that the Operator and Literal ClickGroups have been combined together. I felt this was better looking and took up less space than having them separate. Other than that, these are the same.

### Input, Roll, Reroll
Lastly, we have the input field. It's a bit more styled, but behaves the same for the most part. The only real difference is that it now scrolls to the end as you hit the on screen buttons so you can see what you're entering. Not a big difference, but nice.

The roll button is also just slightly styled with no other change. It now does a press animation when clicked and becomes disabled when the input box is empty.

The reroll button is a new addition. It simply populates the input field with the input string of the roll in the Last Results section. It overwrites anything in the input box, so be careful. This along with the reroll button in history was a feature requested in the discord, though there isn't a github issue to link.

### Known Issues
Right now I can point to one major implementation problem. I tried to make babylonjs work with me, but never could fully succeed. The current method of switching between 3D dice in the dice box and on the game board is to generate 2 separate DiceThrowers, one for each. Then keep them cached and switch between them as requested. This is not ideal. The other option I tried out was to have only one engine loaded at a time and whenever a switch to the other canvas was requested, dispose of the old engine and generate a new one from scratch. This caused stuttering whenever a new engine loaded, which was also not ideal.

I saw info on having multiple render targets on the babylonjs docs, but there were some issues with implementing that solution that I couldn't get past. The first time rendering to a canvas, the geometry was slightly off which caused the dice to not behave properly. Closing the view and reopening it fixed the issue, but I don't know why.

With the current 2 engine setup, this issue doesn't stop the tool from working, it just can cause the page to become a bit heavier than is necessary. I did find that switching between the 2 canvases as the engine was loading would cause the engine creation to fail, so I made it so while either engine is loading, the toggle switch is temporarily disabled. The load takes less than 1 second, so shouldn't be an issue.

Another minor issue is the lack of i18n. I thought about following @SuikaXhq's lead and adding the i18n strings, but I know I couldn't do it as well as they did in #1518. I can barely come up with a good name for a variable let alone that many! I can try doing it if so requested.

### Demo

https://github.com/user-attachments/assets/b07a107f-59b3-4c90-b483-19ddc6e27e84


SuikaXhq:
i18n names I made are actually a bit casual as long as they are clear enough. I can try to replace your labels into i18n strings and submit a PR to you in soon.

client/src/game/systems/dice/index.ts
```diff
@@ -41,12 +41,10 @@ class DiceSystem implements System {
...
-    async load3d(): Promise<void> {
-        if (!diceState.raw.loaded3d) {
-            const env = await getDiceEnvironment();
-            await env.loadDiceEnv();
-            $.loaded3d = true;
-        }
+    async load3d(canvas?: HTMLCanvasElement): Promise<void> {
+        const env = await getDiceEnvironment();
+        await env.loadDiceEnv(canvas);
+        $.loaded3d = true;
```
Kruptein, line 47: loaded3d no longer has any value

client/src/game/ui/tools/DiceTool.vue
```diff
@@ -11,20 +13,52 @@ import { DxHelper } from "../../systems/dice/dx";
 import { diceState } from "../../systems/dice/state";
 import { diceTool } from "../../tools/variants/dice";
 
-const showHistory = ref(false);
+const { t } = useI18n();
+
+// const limitOperatorOptionNames = [t('game.ui.tools.DiceTool.limit_operator_options.keep'), t('game.ui.tools.DiceTool.limit_operator_options.drop'), t('game.ui.tools.DiceTool.limit_operator_options.min'), t('game.ui.tools.DiceTool.limit_operator_options.max')]
+// const selectorOptionNames = ["=", ">", "<", t('game.ui.tools.DiceTool.selection_option_names.highest'), t('game.ui.tools.DiceTool.selection_option_names.lowest')]
```
Kruptein, line 19: What's the deal with these?

client/src/game/ui/tools/DiceTool.vue
```diff
@@ -70,157 +112,584 @@ function addSelector(selector: (typeof DxConfig.selectorOptions)[number]): void
...
+    const parts = roll.parts;
+
+    let content = "";
+
+    for (const part of parts) {
+        content += part.input ?? "";
+    }
+    input.value = diceState.raw.systems!["2d"].parse(content);
+}
+function populateInputFromHistoryIndex(index: number): void {
```
Kruptein, line 139: Add spacing

client/src/game/ui/tools/DiceTool.vue
```diff
@@ -70,157 +112,584 @@ function addSelector(selector: (typeof DxConfig.selectorOptions)[number]): void
...
-    const results = await diceTool.roll(
-        inputText.value,
-        use3d.value === "yes",
-        shareResult.value.toLowerCase() as DiceRollResult["shareWith"],
-    );
     clear();
-    diceSystem.showResults(results);
+    awaitingRoll.value = true;
+
+    const getResult = async (): Promise<RollResult<Part>> => {
```
Kruptein, line 150: Why did this need to become a function?

client/src/game/ui/tools/DiceTool.vue
```diff
@@ -70,157 +112,584 @@ function addSelector(selector: (typeof DxConfig.selectorOptions)[number]): void
...
+                        <div
+                            v-for="[index, part] of lastRoll.parts.entries()"
+                            :key="index"
+                        >
+                            <div v-if="lastRoll.result === '-'" style="font-style: italic"> <!-- starting state -->
+                                {{ part.shortResult }}
+                            </div>
+                            <div v-else-if="part.longResult" class="dice-result">
+                                <div class="input">{{ part.input ?? "" }}</div>
+                                <div class="ops">{{ '(' + part.longResult.replaceAll(',', ' + ') + ')' }}</div>
```
Kruptein, line 209: I think it would be nice to add an extra option "Output" with options "short" and "long" or "expanded" or something. Which would toggle the visibility of this element.

Some people care about seeing the individual results, others just want the results and don't like the clutter of all those individual items. Having a simple toggle that can be toggled if you do need the details can be interesting I think. (And at all times have a hover message when you hover over a result show the details as well)

We can make the dice UI element a bit wider to fit 1 third toggle in that options row.

It might also be interesting to make this a markdown aware field so that the various notations for rolls that were kept or dropped also work. (See `systems/chat/md.ts` for an example, we only need a very basic markdown instance without special setups)

client/src/game/ui/tools/DiceTool.vue
```diff
@@ -70,157 +112,584 @@ function addSelector(selector: (typeof DxConfig.selectorOptions)[number]): void
...
+        <div class="drawer-toggle" @click="showRollHistory = !showRollHistory">
+            <div class="toggle-label">{{ t('game.ui.tools.DiceTool.full_history') }}</div>
+            <font-awesome-icon class="toggle-chevron" :icon="showRollHistory ? 'minus' : 'plus'" :title="t('game.ui.tools.DiceTool.full_history_title')" />
+        </div>
+        <div class="drawer-transition-wrapper">
+            <Transition name="drawer-expand">
+                <div v-show="showRollHistory" id="dice-history-drawer" class="drawer"> <!--Make scrollbar conform to border-radius-->
+                    <div
+                        id="dice-history"
+                    >
```
Kruptein, line 232: This can be on 1 line

client/src/game/ui/tools/DiceTool.vue
```diff
@@ -70,157 +112,584 @@ function addSelector(selector: (typeof DxConfig.selectorOptions)[number]): void
...
+                                                        <div class="input">{{ part.input ?? "" }}</div>
+                                                        <div class="ops">{{ '(' + part.longResult.replaceAll(',', ' + ') + ')' }}</div>
+                                                        <div class="value">{{ part.shortResult }}</div>
+                                                    </div>
+                                                    <div v-else-if="part.shortResult === '+' || part.shortResult === '-'" class="operator-result">
+                                                        {{ part.shortResult }}
+                                                    </div>
+                                                    <div v-else class="literal-result">
+                                                        {{ part.shortResult }}
+                                                    </div>
```
Kruptein, line 271: This is completely repeated from the main result box if I'm not mistaken.

Might be worth to move this out to a separate component then (e.g. `DiceResults.vue`)

Kruptein (changes requested):
Mostly small fluff.

I did check it out locally however and the 3d dice: On version, just gives me a big white screen. For some reason the background of the main babylon canvas is no longer transparent.

client/src/game/ui/tools/DiceTool.vue
SuikaXhq: These should be the i18n names in ClickGroup. I commented instead of removing them since you have said there would be an update on it to have the same API as ToggleGroup (like `{label: string; value: T}`) and they could be put into use then.

rexy712:
> I did check it out locally however and the 3d dice: On version, just gives me a big white screen. For some reason the background of the main babylon canvas is no longer transparent.

Hm that's weird and concerning. I don't see this behavior in chromium or firefox. I made a fresh clone of the repo to make sure it wasn't something lingering in my working copy, and I don't see the behavior there either. It should still clear to transparent with the line `scene.clearColor = new Color4(0, 0, 0, 0);`

Is it a white background since connecting to a game or only after loading the 3D dice?

Kruptein:
> > I did check it out locally however and the 3d dice: On version, just gives me a big white screen. For some reason the background of the main babylon canvas is no longer transparent.
> 
> Hm that's weird and concerning. I don't see this behavior in chromium or firefox. I made a fresh clone of the repo to make sure it wasn't something lingering in my working copy, and I don't see the behavior there either. It should still clear to transparent with the line `scene.clearColor = new Color4(0, 0, 0, 0);`
> 
> Is it a white background since connecting to a game or only after loading the 3D dice?

Only after checking the 3d and the box variant worked fine. I checked the clearcolor and that was indeed still fine so unsure.
If you can't reproduce it I'll try to figure it out

client/src/game/ui/tools/DiceTool.vue
rexy712: > I think it would be nice to add an extra option "Output" with options "short" and "long" or "expanded" or something. Which would toggle the visibility of this element.

I agree that giving the option is a nice idea. I really don't want to increase the width of the tool if I can avoid it. Right now the width is set to nicely fit the literal+symbol clickgroup, which looks good. I also don't want to increase height because it can get pretty tall with mutiple drawers open already.

I think some sort of options menu might be nice for this. Keep things that will only be changed infrequently hidden most of the time to cut down on clutter. In the this menu, have separate options for last roll breakdown details and history item breakdown details. This also makes the tool more expandable in the future if more options are added. Obviously keep 3D dice and Share on the main screen since they're going to be used quite often.

> It might also be interesting to make this a markdown aware field so that the various notations for rolls that were kept or dropped also work.

Somehow I didn't even realize the notation was supposed to be markdown lol. I could get that working.

client/src/game/ui/tools/DiceTool.vue
rexy712: It didn't. Just my lack of javascript/typescript experience showing

SuikaXhq:
> I did check it out locally however and the 3d dice: On version, just gives me a big white screen. For some reason the background of the main babylon canvas is no longer transparent.

Tested at commit 5e37a9d, didn't encounter that problem.

Kruptein:
Ok I figured it out. I only could reproduce it on my windows desktop at first and not on my linux laptop, so I thought some strange OS thing.

But it's because on my desktop I have a big screen, which ends up causing the camera's height to end up above the roof of the dicebox

rexy712:
> But it's because on my desktop I have a big screen, which ends up causing the camera's height to end up above the roof of the dicebox, because the camera's position is calculated based on the canvas height.

Ah I see. I don't really like the camera height being based off one dimension of the canvas like that anyway. I think I'll just make them 2 constants. I was trying to be more clever than was warranted.

December is my busy season so progress will be slower. I'll still work on it when I have time.

PR #1583, opened 2025-04-07
feat(Mods): Mod add/remove UI

Kruptein:
Back in [release 2023.3](https://www.planarally.io/blog/release-2023.3/#mods) experimental features were added to start having some framework for mod support in PA.

That release added the ability to store data on the server and access to a subset of PA APIs.

There are many things that are still lacking in the mod API, it's experimental after all, but one of the biggest problems was that the only way to add a mod was to have the server owner add it to a specific folder on the server and it would be loaded for all users who use the server.

I've now taken the time to actually make this process something that is more granular and no longer a server-owner-only flow.

This PR adds a variety of mod related changes, but the most prominent one is the ability for a DM to upload mods that are then enabled for all players in the campaign. These mods are still stored in the server's `static/mods` folder, but now use a specific folder structure to ensure that mods with a same name or different versions of the same mod, are properly separated.

To upload a mod, you can go to the DM settings new mod tab.

Additionally the mod API (mostly the datablock APIs) has updated quite a bit as well. A new `useShapeDataBlock` hook has been added that simplifies a bunch of logic in mod components.

All mods in the sample mod repository (https://github.com/Kruptein/planarally-mods) have been updated to the changes in this PR.

_do note that the mod API remains experimental_


develroo:
Good news. Thanks.

PR #1603, opened 2025-05-31
bug(Loading): Fix load whitescreen

Kruptein:
Before we open the game proper we already create the socket to the server so that we can already start receiving data while loading the UI.

Recently an extra request was added before opening the UI to fetch the mod list, which ended up causing a race-condition where sometimes the socket data would already arrive and try to add layers to the board when the UI wasn't ready yet, causing a whitescreen with no game.

I've changed this loading strategy so that the mod loading no longer blocks this process.


Kruptein:
The linting error is on an unused export function, which will be addressed in a follow-up cleanup PR

PR #1608, opened 2025-06-11
feat(Stats): Add stats collection

Kruptein:
## Intro

A question I often receive is how many users PA actually has and I can never actually answer this or even estimate this.

There is no official PA server that I can base myself on for some estimates, I only have some vague numbers like github stars or discord users, but those don't really say anything about actual users and if they're any indication at all they're more likely to represent server owners than players.

In order to get a feeling for what's actually going on out there I'm introducing a stats collection thingy on the server.

## Which stats

Although it would be nice to get some metrics into usage of various features of PA to see if things are actually used or not, this PR will focus on just 2 metrics. (And I don't expect other metrics being added anytime soon)

- Server total size: Get the total number of users and campaigns
- Server active use: Get the number of unique users and campaigns that interact with the server on a regular basis (e.g. Server X has 12 unique users active in the month of August)

## Where are the stats collected

Each local server will collect them locally and the goal is to ping this data on a weekly* basis to https://stats.planarally.io (yes there is nothing to see here)

Note that a lot of this will be configurable (see last section)

\* Initially the default will be set to daily to get a feeling for how it's behaving and catch early issues.

## How are the stats used

The goal is to get a feeling of actual use. Initially the stats will just exist in the database of each server as well as the stats server with no UI.

The goal is to provide a UI on the stats server itself that will show global use as well as provide admins a stat view of their own server.

## Privacy

The data that the stats server receives will consist of hashed IDs. They are hashed with a pepper that is uniquely generated on the local servers. It's also all just part of PA so it's opensource!
No other user info is provided (e.g. usernames, email addresses, ...).

Similarly the only info that I track about the server sending the data is a uuid that the server generates itself that is only used for sending the stats and the PA version in use, no other information (e.g. IP address) is stored.

## Configuration

This stats collection module comes with a bunch of configuration options for the server owner.
See the [PA docs](https://deploy-preview-191--planarally.netlify.app/server/management/configuration/#stats) for the details.

You'll be able to disable stats collection entirely, you can also enable it but disable the export to the PA stats server. Also the frequency of stats exporting will be adjustable as well as the endpoint that the stats are sent too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ They will be removed in a future release though.
     -   Upload/Remove mods for a specific campaign
 -   Asset create folder/upload file/remove buttons to the in-game asset browser
 -   [server] Email configuration setup
+-   [server] Anonymous stat collection
+    -   Sends usage stats about number of campaigns and users to stats.planarally.io
+    -   Identifiers are anonymized, no personal info is included
+    -   Can be disabled
 
 ### Changed
 
--- a/server/src/api/common/rooms/create.py
+++ b/server/src/api/common/rooms/create.py
@@ -1,5 +1,6 @@
 from typing import Optional
 
+from .... import stats
 from ....db.create.floor import create_floor
 from ....db.db import db
 from ....db.models.location import Location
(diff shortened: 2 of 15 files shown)


10leej:
This is a welcome change and i support it as someone how had to manually count usernames in the past.

PR #1635, opened 2025-07-14
bug(Initiative): Fix issues related to round/turn handling

rexy712:
This PR fixes a number of smaller issues in the handling of turn changes and round changes. I'll address each individually:

## Effect Timers
When going to a previous turn in initiative, the effects attached to an entry would decrement as though you had gone to the next turn. I decided to add navigation direction to the calls to `setTurnCounter` in the client and `update_initiative_turn` in the server. It was otherwise impossible to tell which direction (forward or backward) the initiative was moving when only 2 entries were present. This information allows us to decrement the effect counters of an entry only when going forward.

I also thought that *incrementing* the effect counters when going in reverse seemed desirable. Imagine accidentally advancing and going back causing the effect timers to be inaccurate.

https://github.com/user-attachments/assets/eeaa2160-55e3-46ac-ab40-6f1a77a0f497

## Invalid Turn Index

When removing a shape from the beginning of the initiative list, the turn counter would enter an invalid state. This was a straight forward fix of index checking.

https://github.com/user-attachments/assets/d9c0fdcc-84da-4aca-bb1b-704157498351

## Negative Round Counter

The round counter could go below 0 by pressing the previous turn button repeatedly. I considered this as an invalid state, though the logic had no problem handling it. I believe that round 0 is a valid round as it could be treated as a 'surprise round'. You could make the same argument for negative rounds being additional surprise rounds, but this seems like a rather hacky way to use the tool.

https://github.com/user-attachments/assets/5275146a-b86f-42d4-b704-412472395d57

## Entry Removal Should Go to Next Turn

When removing an entry from initiative, I believe that one would typically want to go to the next entry in the list rather than the one preceding. Usually when I remove an entry, it has left combat due to fleeing or death and the next entry would immediately take its turn. The current behavior is to go to the previous entry. I believe this would cause issues with effect timers and add an extra step to finding the next entry to act.

### DIFF
--- a/server/src/api/socket/initiative/__init__.py
+++ b/server/src/api/socket/initiative/__init__.py
@@ -288,12 +288,10 @@ async def remove_initiative(sid: str, data: str):
             )
         elif shape_turn == location_data.turn:
             # In this case we just want to proceed to the next actor in line as if we normally would advance
-            next_turn = shape_turn
             if shape_turn == len(json_data) - 1:
-                next_turn = 0
                 location_data.turn = 0
                 if len(json_data) > 1:
-                    location_data.round = location_data.round + 1
+                    location_data.round += 1
                     await _send_game(
                         "Initiative.Round.Update",
                         location_data.round,
@@ -301,7 +299,9 @@ async def remove_initiative(sid: str, data: str):
                     )
             await _send_game(
                 "Initiative.Turn.Update",
-                InitiativeTurnUpdate(turn=next_turn, direction=InitiativeDirection.FORWARD, processEffects=True),
+                InitiativeTurnUpdate(
+                    turn=location_data.turn, direction=InitiativeDirection.FORWARD, processEffects=True
+                ),
                 room=pr.active_location.get_path(),
             )
         location_data.data = json.dumps([initiative for initiative in json_data if initiative["shape"] != data])
@@ -362,16 +362,16 @@ async def update_initiative_turn(sid: str, raw_data: Any):
 
     location_data: Initiative = Initiative.get(location=pr.active_location)
     json_data = json.loads(location_data.data)
-    data_len = len(json_data)
+    total_turn_count = len(json_data)
 
     turn = data.turn
     process_effects = data.processEffects
 
-    if turn < 0 or turn >= data_len:
+    if turn < 0 or turn >= total_turn_count:
         logger.warning("Provided turn is out of bounds.")
         return
 
-    db_turn_valid = 0 <= location_data.turn < data_len
+    db_turn_valid = 0 <= location_data.turn < total_turn_count
 
     if db_turn_valid:
         shape = Shape.get_or_none(uuid=json_data[location_data.turn]["shape"])
@@ -397,7 +397,7 @@ async def update_initiative_turn(sid: str, raw_data: Any):
                         try:
                             turns = int(effect_turns)
                             if turns <= 0 and data.direction == InitiativeDirection.FORWARD:
-                                json_data[entry]["effects"].pop(i)
+                                effect_list.pop(i)
                                 continue
                                 # Element removed, do not increment
                             else:
--- a/server/src/api/socket/initiative/__init__.py
+++ b/server/src/api/socket/initiative/__init__.py
@@ -277,11 +277,9 @@ async def remove_initiative(sid: str, data: str):
         location_data = Initiative.get(location=pr.active_location)
         json_data = json.loads(location_data.data)
 
-        location_data.data = json.dumps([initiative for initiative in json_data if initiative["shape"] != data])
-
         shape_turn = next(i for i, v in enumerate(json_data) if v["shape"] == data)
 
-        if shape_turn < location_data.turn or shape_turn == len(json_data) - 1:
+        if shape_turn < location_data.turn:
             location_data.turn -= 1
             await _send_game(
                 "Initiative.Turn.Set",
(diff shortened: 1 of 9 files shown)


server/src/api/socket/initiative/__init__.py
```diff
@@ -290,14 +288,23 @@ async def remove_initiative(sid: str, data: str):
...
+                    location_data.round = location_data.round + 1
+                    await _send_game(
+                        "Initiative.Round.Update",
+                        location_data.round,
+                        room=pr.active_location.get_path(),
+                    )
             await _send_game(
                 "Initiative.Turn.Update",
-                location_data.turn,
+                InitiativeTurnUpdate(turn=next_turn, direction=InitiativeDirection.FORWARD, processEffects=True),
```
Kruptein, line 304: This can remain `location_data.turn` for the turn, we don't need the `next_turn` variable as it always has the same value as `location_data.turn`

server/src/api/socket/initiative/__init__.py
```diff
@@ -290,14 +288,23 @@ async def remove_initiative(sid: str, data: str):
...
         elif shape_turn == location_data.turn:
             # In this case we just want to proceed to the next actor in line as if we normally would advance
-            # And thus process effects
-            location_data.turn -= 1
+            next_turn = shape_turn
+            if shape_turn == len(json_data) - 1:
+                next_turn = 0
+                location_data.turn = 0
+                if len(json_data) > 1:
+                    location_data.round = location_data.round + 1
```
Kruptein, line 296: Nitpick:
```suggestion
                    location_data.round += 1
```

server/src/api/socket/initiative/__init__.py
```diff
@@ -349,17 +356,22 @@ async def change_initiative_order(sid: str, raw_data: Any):
...
 @sio.on("Initiative.Turn.Update", namespace=GAME_NS)
 @auth.login_required(app, sio, "game")
-async def update_initiative_turn(sid: str, turn: int):
+async def update_initiative_turn(sid: str, raw_data: Any):
+    data = InitiativeTurnUpdate(**raw_data)
     pr = game_state.get(sid)
 
     location_data: Initiative = Initiative.get(location=pr.active_location)
     json_data = json.loads(location_data.data)
+    data_len = len(json_data)
```
Kruptein, line 365: `data_len` is a correct name, but maybe `number_of_turns`/`total_turns`/`turn_count` or something similar is a clearer name for the actual use of the variable.

server/src/api/socket/initiative/__init__.py
```diff
@@ -374,30 +386,35 @@ async def update_initiative_turn(sid: str, turn: int):
             return
 
         with db.atomic():
-            next_turn = turn > location_data.turn
+            if process_effects:
+                entry = location_data.turn if data.direction == InitiativeDirection.FORWARD else turn
+                effect_list = json_data[entry]["effects"]
+                i = 0
+                while i < len(effect_list):
```
Kruptein, line 393: I prefer keeping the original for loop, you're now using 5 lines to do the same thing that was being done in 1

server/src/api/socket/initiative/__init__.py
```diff
@@ -374,30 +386,35 @@ async def update_initiative_turn(sid: str, turn: int):
...
+                    _effect = effect_list[i]
+                    effect_turns = _effect["turns"]
+                    if effect_turns is not None:
+                        try:
+                            turns = int(effect_turns)
+                            if turns <= 0 and data.direction == InitiativeDirection.FORWARD:
+                                json_data[entry]["effects"].pop(i)
```
Kruptein, line 400:
```suggestion
                                effect_list.pop(i)
```

Kruptein (changes requested):
Great changes all around, my only comments are stylistic in nature and not about the logic.

The only thing I'm still somewhat on the fence about is the change to the round counter.

I feel like either we disallow going back all together, which means also no round 0, or we just let the user do whatever wacky thing they want to do.

server/src/api/socket/initiative/__init__.py
rexy712: So the reason I changed this to a while loop is because when an item is popped it seemed to cause issues with the iterators that the for loop was using. On line 401, the *i* variable needs to not update because an element was removed. If *i* were to increment there, one element in the list would be skipped. In all other cases, the index should still increment.

server/src/api/socket/initiative/__init__.py
rexy712: True, not sure why it's there. Probably from testing something and forgot to remove it all the way

server/src/api/socket/initiative/__init__.py
Kruptein: That's why the iteration actually is in reverse (the `[-1:]`), though the pop should actually have been done by subtracting i from the length instead of using i directly.

An alternative is to do `for i in range(data_len - 1, 0, -1)` but then you still need to grab the element by indexing

rexy712:
> The only thing I'm still somewhat on the fence about is the change to the round counter.

Fair, it was a feeling I had that it was something undesirable. I can change this to reallow negative numbers if desired.

> Sorry for the delay on the review by the way. I was on holiday and my main desktop's cpu died when I came back :D

No problem at all! I've been working on a reskin of the initiative tracker that depends on this change, so the delay has given me time to reevaluate some aspects and improve it overall before I try to make it PR ready.

PR #1640, opened 2025-08-06
feat(Admin): Move admin dashboard into main app

Kruptein:
The old admin dashboard hasn't been updated in a while and is unwieldy to use for most users.

The idea was to shield it as much from the outside world as possible, but in the process it was mostly shielded from users as well. It also made it that I didn't quickly update the code as it was pretty much a separate application.

This PR removes all code related to the old admin client and also removes the API system there used to be. If the latter is something you actively used, do let me know.

It no longer uses authentication tokens that you have to generate, but instead a new entry in the config's general section called "admin_user" which should be the name of the user that you want to have root admin powers.

The goal is to have a proper permission system later down the line that the admin user can define and which will be tracked in the database.

For now it's just the main admin user that will be able to open the admin panel, using an extra "admin" section in the dashboard that will appear. It will bring the user to a separate page with currently just 2 entries:
- users
- campaigns

This should be the same data that was previously available in the separate admin app.

The goal here is to also expand a bunch on this:
- show the stats we've been collecting since the last release
- the aforementioned permission stuff
- changing some of the config settings
- adding new users directly
- ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ tech changes will usually be stripped from release notes for the public
 -   Admin panel that can be accessed from the dashboard
     -   Currently limited to the main admin_user as configured in the server's config
     -   Shows roughly same content as the original separate admin client
-    -   Also shows total number of users and a quick user create button
+    -   Also shows total number of users/campaigns and a quick user create button
 
 ### Changed
 
--- a/client/src/admin/Campaigns.vue
+++ b/client/src/admin/Campaigns.vue
@@ -43,6 +43,10 @@ const filteredCampaigns = computed(() => {
                     <div class="fullWidth">No campaigns match the current filter.</div>
                 </template>
             </div>
+
+            <div id="action-bar">
+                <div>Total: {{ campaigns.length }}</div>
+            </div>
         </div>
     </section>
 </template>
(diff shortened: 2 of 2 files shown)


Kruptein:
> Hmm I added this to my config.toml ? Is that correct ?
> 
> ```
> [admin_user]
> admin = true
> ```

No, the description says

> a new entry in the config is added under the general section "admin_user" which should be the name of the user that you want to have root admin powers.

So under the "general" section in your config add an entry "admin_user" with as the value a username.

e.g.

```toml
[general]
admin_user = "someusername"
```

I've updated the text a bit to make it clearer that it should go under general, but how did you miss the username part :p

develroo:
> > Hmm I added this to my config.toml ? Is that correct ?
> > ```
> > [admin_user]
> > admin = true
> > ```
> 
> No, the description says
> 
> > a new entry in the config is added under the general section "admin_user" which should be the name of the user that you want to have root admin powers.
> 
> So under the "general" section in your config add an entry "admin_user" with as the value a username.
> 
> e.g.
> 
> ```toml
> [general]
> admin_user = "someusername"
> ```
> 
> I've updated the text a bit to make it clearer that it should go under general, but how did you miss the username part :p

Ahhh I tried it without quotes and it broke. hehe. Thanks.

PR #1663, opened 2025-09-13
Add OIDC implementation with authentication and configuration support

JatinVasman:
Implements OpenID Connect authentication allowing users to login with external providers.
Features:
OIDC authentication flow with multiple provider support
Account linking via email for existing users
Configurable via environment variables
Added login UI with provider buttons
Changes:
Backend: OIDC auth module, API endpoints, configuration
Frontend: Updated login page with OIDC buttons and callback handling
Config: Environment template and Docker compose setup
No breaking changes - OIDC is optional and works alongside existing auth.

### DIFF
--- a/server/src/auth/oidc.py
+++ b/server/src/auth/oidc.py
@@ -45,11 +45,11 @@ class OIDCAuth:
             return None
 
         try:
-            # Enforce HTTPS for security - reject HTTP completely
+            # Enforce HTTPS for security - only allow secure protocols
             domain = self.config.domain
-            if domain.startswith('http://'):
-                # Reject insecure HTTP protocol completely
-                logger.error(f"Insecure HTTP protocol rejected for OIDC domain: {domain}. Only HTTPS is allowed.")
+            if '://' in domain and not domain.startswith('https://'):
+                # Reject any non-HTTPS protocol
+                logger.error(f"Insecure protocol rejected for OIDC domain: {domain}. Only HTTPS is allowed.")
                 return None
             elif not domain.startswith('https://'):
                 # Add HTTPS if no protocol specified


server/src/routes.py
```diff
@@ -18,22 +22,73 @@
...
+    # USERNAME/PASSWORD CONFIGURATION with fallback safety
+    # If username/password is disabled BUT OIDC is not available, enable username/password as fallback
+    username_pass_should_be_disabled = (
+        not config.general.username_password_enabled and 
+        oidc_fully_configured  # Only disable if OIDC is actually available
+    )
+    
+    if username_pass_should_be_disabled:
+        data = data.replace(b'name="PA-username-pass" content="true"', b'name="PA-username-pass" content="false"')
+    elif not config.general.username_password_enabled and not oidc_fully_configured:
```
Kruptein, line 46: This is now checked twice, both in the condition on line 39 and here

server/src/routes.py
```diff
@@ -18,22 +22,73 @@
...
+        # Fallback activated: username/password was disabled but OIDC is not available
+        # Keep username/password enabled to prevent users from being locked out
+        pass  # Meta tag stays as content="true"
+    
+    # MAIL CONFIGURATION
+    if config.mail is None or not config.mail.enabled or not config.mail.host or not config.mail.port or not config.mail.default_from_address:
```
Kruptein, line 52: host, port and the default email cannot be empty if `config.mail` exists so checking them is unnecessary

server/src/routes.py
```diff
@@ -18,22 +22,73 @@
...
     if not config.general.allow_signups:
         data = data.replace(b'name="PA-signup" content="true"', b'name="PA-signup" content="false"')
-    if not config.mail or not config.mail.enabled:
+    
+    # OIDC CONFIGURATION - Check first to determine if it's available
+    oidc_fully_configured = (
+        config.oidc.enabled and 
+        config.oidc.client_id and 
+        config.oidc.client_secret and
+        bool(config.oidc.domain or config.oidc.authorize_url)  # Either domain for discovery OR direct URLs
```
Kruptein, line 34: These should just become mandatory in the config. If you specify an oidc config, it should contain all the data required.

For the domain vs authorize_url we probably should ideally check that in a sort of post-validation step in the config manager itself, but I'm ok with keeping it here for now

server/src/routes.py
```diff
@@ -18,22 +22,73 @@
...
+    template = "admin-index.html" if admin_api else "index.html"
     with open(FILE_DIR / "templates" / template, "rb") as f:
         data = __replace_config_data(f.read())
         return web.Response(body=data, content_type="text/html")
 
 
-async def root_dev(request):
-    port = 8080
+async def root_dev(request, admin_api=False):
+    port = 8081 if admin_api else 8080
```
Kruptein, line 91: Revert these

server/src/routes.py
```diff
@@ -71,8 +130,23 @@ async def root_dev(request):
...
+admin_app.router.add_static(f"{subpath}/static", STATIC_DIR)
+admin_app.add_subapp("/api/", api_app)
+
 TAIL_REGEX = "/{tail:(?!api).*}"
 if "dev" in sys.argv:
     main_app.router.add_route("*", TAIL_REGEX, root_dev)
+    admin_app.router.add_route("*", TAIL_REGEX, partial(root_dev, admin_api=True))
 else:
     main_app.router.add_route("*", TAIL_REGEX, root)
+    admin_app.router.add_route("*", TAIL_REGEX, partial(root, admin_api=True))
```
Kruptein, line 152: Revert these

Kruptein:
I'm going to close this in favour of #1701, which is based on your changes (and credits you in the changelog).

PR #1688, opened 2025-10-26
chore(Deps): Upgrade pydantic

Kruptein:
This PR upgrades pydantic from v1 to v2. Pydantic is a library the server uses for data modelling/validation and in particular at the API treshold to ensure data is in the formats the server expects.

A while back v2 came out with some backwards incompatible changes and I now found the time to actually get around upgrading.

This _should_ have no impact on users, but given the breaking nature of the upgrade I might have missed some things here and there.


develroo:
I assume this will be rolled into the docker image?

Issue #394, opened 2020-06-04
[FEAT] Origin Mark


chisaipete:
**Is your feature request related to a problem? Please describe.**
When building location maps, it would be nice to have a DM visible mark for where the (0,0) point is in the map.  This allows you to be able to center assets to the map without relying on markers or spawn location adjustments. 

**Describe the solution you'd like**
Something similar to #343 would work fine, perhaps just a little circle or pip on the graph at the intersection.

dthv:
Since this will probably appear on the grid layer: A setting to toggle visibility would be nice, as one might need the mark on locations with invisible grid. Of course, even if the grid is displayed, one might not want to see the origin mark.

Another possibility would be to have a non-permanent mark, only displaying on certain user action. E.g. making Ctrl+Shift+0 the hotkey to mark origin.

PR #1712, opened 2025-11-24
Add Origin Marker

Craftidore:
Adds an origin circle as requested in #394 .

image

This is an initial functional implementation to make sure I'm on the right track here.
I stole most of the implementation from `shapes/variants/circle.ts`.

Adjustments needed before it's ready to merge:

- DM Setting/Location Setting for whether to show origin
- DM Setting/Location Setting for origin circle color (and radius?)
- DM Setting/Location Setting for whether to show Players the origin circle

Questions for you about preferred implementation:

- Should origin circle be on its own layer, or is it fine to put it in the grid layer?
  - I'm assuming it does need to be its own layer, since the entire grid layer is hidden upon disabling the grid.
- This is my first time working with an html canvas, so some questions:
  - How concerned should I be about resetting the ctx to its initial state after finishing my drawing? For example, I'm remembering then resetting fillState after drawing, but that's redundant if your code doesn't assume that.
  - When you call a draw function, after calling a setTransform, and that draw function calls setTransform then resetTransform, does that preserve the initial transform like a push/pop action, or does it full reset the transform? If the latter, how should I handle this instead?
- Do you want the grid to bleed through the origin marker?

### DIFF
--- a/client/src/fa.ts
+++ b/client/src/fa.ts
@@ -34,7 +34,6 @@ import {
     faCog,
     faCogs,
     faCommentDots,
-    faLocationDot,
     faCut,
     faDiceSix,
     faDoorClosed,


Craftidore:
Ok, another thought: Maybe the icon should be bigger. If you zoom out the last or second to last zoom level, you can't see it at all, which is a problem. If I up the size to 40, it becomes barely visible at max zoom level.
image
That makes it kinda large at lesser zoom levels though. I dunno how intrusive that feels.
image

Kruptein:
By default shapes size along with the zoom level. You can however pass a third parameter to the draw function to set a custom width/height which will be fixed at all zoom levels. You could try and see how that feels.

rexy712:
> I'm personally not really sure if a circle is super clear as the origin symbol, some specific icon might be clearer. 

I would advocate for one of these icons, but that's just my taste. These are fontawesome location-crosshairs and location-dot.
image
image

> Do you want the grid to bleed through the origin marker?

My vote would be for yes. Making the origin mark slightly transparent in general would be nice, so that the piece of a map underneath isn't entirely obscured by it.

PR #1728, opened 2026-01-10
Make snapping of asymmetric shapes on square grid better align to the grid

rexy712:
When you have an asymmetric shape on a square grid, the snapping behavior doesn't act as one would expect it to. Instead of snapping the corners to the grid, it snaps each of the shape's corners to the midpoint of a grid edge. This makes the shape seem misaligned to the grid and difficult to resize and realign.

The reason for this behavior is that grid size inference is based on a single variable which is set to the max(width, height) of the shape. When a shape has equal width and height, this works as expected with the shape corners snapping to the corners of the grid. When the shape width and height that are not equal (more specifically when a shape has an even length width and odd length height or vice versa), the snapping is misaligned.

This PR gives each shape a size_x and size_y variable for grid alignment so that an asymmetric shape can adjust snapping width and height to better fit the grid. This behavior only affects square grids. Hex grids are left untouched.

Below are videos that somewhat demonstrate the issue and the solution. The first is the current behavior, the second is behavior with this PR applied.

https://github.com/user-attachments/assets/1d6841e0-de9e-4d2c-b8b9-9ecd466b0cce

https://github.com/user-attachments/assets/e4f83737-fe3d-4927-94dc-b512da689390

*Note*: I did this work before the intermediate shape data representation update and forgot to ever create a PR. Hopefully there isn't a glaring issue I didn't account for when porting it over.


client/src/game/systems/properties/types.ts
```diff
@@ -19,7 +21,7 @@ export interface ShapeProperties {
     isDefeated: boolean;
     isLocked: boolean;
     // grid related
-    size: number; // if 0, infer size
+    size: ShapeSize; // if 0, infer size
```
Kruptein:
```suggestion
    size: ShapeSize; // if x and y are 0, infer size
```

Kruptein (changes requested):
Nice, this is something I've been wanting to do in the past as well. I gave up when I wanted to handle the hex side, but it's probably smart to keep that separate like you did.

Issue #1734, opened 2026-01-19
[BUG] Players Always Sent to First Floor


Craftidore:
**Describe the bug**

When you send a player to another floor via moving their characters to a different location and spawn location, the players are sent to the bottom floor instead of the floor of the spawn location their character was sent to.

**To Reproduce**
1. Create new campaign
2. Create new location called "second location"
3. Add some text to "second location" ground floor and a spawn location called "ground entrance".
4. Create new floor in "second location" called "floor 2"
5. Add some different text to floor 2.
6. Add a spawn location to floor 2 called "floor 2 entrance"
7. Add a asset shape to start location.
8. Connect via another browser as a player.
9. Give the player full access rights to the asset shape.
10. Right click the asset shape as GM and send to "second location" and pick the "floor 2 entrance" spawn location.
11. The player will be stuck viewing the ground floor event though their token is on floor 2.

**Expected behavior**

The player should be sent to the same floor that their token was sent to, even when that's not the ground floor.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mix of Windows and Fedora Linux
 - Browser Mix of Vivaldi, Chrome, Brave
 - Planarally Version: 2025.3

**Additional context**

I spent like 40 minutes at the start of the session trying to figure out what was happening.

develroo:
@Kruptein 

Actually, this seems a more serious than that. It seems the player view is being moved, but the token is not.  This is a definite bug, as I know this used to work before. 

https://github.com/user-attachments/assets/1faed1f5-9bf1-41a3-bc09-7af930a80e71

PR #1739, opened 2026-01-20
Fix teleports and player location changes

rexy712:
While looking into https://github.com/Kruptein/PlanarAlly/issues/1734 I couldn't get teleports to work at all. Looking back at the issue, I noticed they were using version 2025.3. Once I switched to that, the teleports worked as expected again. Something happened between then and the current dev branch, so I decided to try tracking it down.

This PR adds missing parameters to the Location.Change network calls and fixes a check that wasn't functioning in server/src/api/socket/shape/init.py.


rexy712:
> This doesn't actually fix the teleport bug itself right?

No, this is just to get the teleports working before looking into that.

PR #1745, opened 2026-01-21
Initiative: enhance group behavior

### LINKED ISSUE #1743: [FEAT] Non-Group Initiative Slots Identify Whole Group

I regularly add enemies which are part of a group to initiative as separate entities. In Pathfinder 2e, it's generally suggested to *not* run groups in the same initiative slot unless it will really help with fight complexity, as doing so can make encounters way swingier than the system intends. When I do so, the tokens in question are not differentiated in the initiative tracker and hovering over the initiative slot highlights every token in that group.

When grouped entities are added to initiative as individual entities, their group id should show as part of their initiative icon, and when such an initiative icon is hovered over, only the specific token that has that initiative slot should be highlighted on the canvas.

Lately I've been adding a 10-turn effect to each token called `#1` and `#2` etc to the initiative slots. This is inconvenient, but workable.

**Additional context**

Image

rexy712:
Consider this a continuation of the [UI rework PR](https://github.com/Kruptein/PlanarAlly/pull/1659). This potentially resolves #1743.

This PR adds a badge to the corner of an initiative entry if the corresponding shape has its own group badge enabled.

This PR also fixes behavior of highlighting all shapes in a group when the initiative entry is not marked as a group. It will now highlight only the corresponding shape when not in group mode, and it will highlight the whole group when in group mode.

https://github.com/user-attachments/assets/c2264c79-17a3-47db-80d4-a6ad319db51a

*Known Issues*: Badge strings are not reactive. If you change the character set or renumber the group entries, the badges in the initiative tracker will not update until a refresh is done.

### DIFF
--- a/client/src/game/ui/initiative/Initiative.vue
+++ b/client/src/game/ui/initiative/Initiative.vue
@@ -258,7 +258,7 @@ function toggleHighlight(actor: InitiativeData, show: boolean): void {
 }
 
 function isGroupMember(actor: InitiativeData): boolean {
-    if (actor.isGroup) return false
+    if (actor.isGroup) return false;
     if (actor.localId === undefined) return false;
     const groupId = groupSystem.getGroupId(actor.localId);
     if (groupId === undefined) return false;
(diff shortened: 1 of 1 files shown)


develroo:
Cool, I got caught out by that and ended up adding them one at a time. 

Thanks!

PR #1747, opened 2026-01-22
Fix snap wall collision #1735

### LINKED ISSUE #1735: [BUG] Tokens Can Snap Through Walls
**Describe the bug**

When walls are created which don't directly line up to the grid, while tokens can't move through the wall, they can grid-snap through the wall when released.

**To Reproduce**
1. Create a irregular shape with the polygon tool (not closed), as a movement blocking shape.
2. Create a basic token
3. Move the basic token so its center is over the edge of a square mostly blocked by the shape.
4. When mouse is released, basic token snaps through the wall, and if it was a player-controlled token gets trapped.

See screenshots

**Expected behavior**

The token should either not snap or snap as close to the center of the square as it can without moving through a wall.

**Screenshots**

Image

Image

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Vivaldi
 - Version: Planarally 2025.3

**Additional context**

Current workarounds:

- Rescue any tokens that players accidentally move through a wall.
- Create a separate set of walls for vision-blocking and movement blocking, where the movement blocking walls are all on grid lines and the vision-blocking ones actually follow the walls.

Craftidore:
Fixes #1735 .

Possible caveats regarding the code:

- I had to change `snapToGrid` interface to take a boolean for if the shape is on the token layer.
- I'm reusing the existing logic from drag.ts's `calculateDelta`, which I did not bother trying to understand.
  - I got correct behavior from shrink=true and incorrect behavior from shrink=false. I don't understand the differences between these two.

### DIFF
--- a/client/src/game/shapes/shape.ts
+++ b/client/src/game/shapes/shape.ts
@@ -231,7 +231,7 @@ export abstract class Shape implements IShape {
         return this._visionPath;
     }
 
-    onLayerAdd(): void { }
+    onLayerAdd(): void {}
 
     // POSITION
 


Craftidore:
> Why did you decide to only apply this to the token layer?
> All other snapping behaviour in PA works consistently on every layer (iirc), so it feels very odd to me to have this one specific case work differently.

Wall collisions (ordinarily) only occur on the Token Layer. When I'm moving around traps and notes in the GM layer or lights in the FoW layer or map-related set dressing in the Map layer, I still want those things to snap to the grid for convenience, but I don't want them to run into walls.

Issue #500, opened 2020-08-16
[FEAT] Allow assets to be stored and loaded from an S3 store


donicrosby:
**Is your feature request related to a problem? Please describe.**
Allowing the server to store and load assets from an S3 bucket would help server admins to not have to worry about storing large amounts of files on the web server and reduce the cost to run an instance on a cloud provider or low storage home server. This would also allow users to host the images on a self-hosted Minio instance pointing to their bulk storage without the need to setup an NFS or SAMBA mount.

**Describe the solution you'd like**
The Boto3 library is a popularly used library for working with S3 buckets, the license is compatible with the project and does a lot of the heavy lifting for the developer. 
In the configuration file you could add an option for storage backend to be either local or s3. If the backend is set to S3 have a section to set the hostname (if blank Boto assumes you're using AWS), and your Access Key and Access Secret Key.

**Describe alternatives you've considered**
Allow the user to setup a way to store their assets remotely so that they don't have to host a lot of storage on the web server.

**Additional context**
None

Kruptein:
I can see this being useful, but I'm going to be upfront and say that this is not going to be something I'm going to pick up in the near future, because there are some other more fundamental things I wish to prioritize first.

donicrosby:
That's fine, I may implement it when I have time and make a pull request if that's alright with you.

Kruptein:
Absolutely, I do need to make some work of starting to document stuff though :D

10leej:
Can you mount the S3 store with SSHFS? You can do that with google cloud. I haven't looked into storage though since my server right now only has 10GB of data on it.

PR #1802, opened 2026-03-07
feat(Assets): Add remote asset storage option

### LINKED ISSUE #1725: [FEAT] Object Storage Support
I would like to use remote object storage, it would allow me better control.

an option for S3 and logic for upload assets

S3fs fuse and local storage neither option works very well for me and since i plan to move to k8s local storage is no go so is s3fs, i could create Persistent volume but excepensive and wlel object storage is a better way for this sort of stuff anyway

Kruptein:
This PR adds the option to configure a S3-compatible storage remote for user uploaded assets. This can alleviate some costs or maintenance on the server host.

This comes with some configuration changes, a PR to the docs will be opened later today (likely) with specifics on that front. I'll update this description with a link once that's done.

By default without any config changes[^1], the original local storage approach will be used and stored in `/static/assets/` (relative to the server folder).

You'll however now be able to instead refer to a S3-compatible host by providing some URL info (e.g. bucket name etc) as well as 2 env variables to provide secrets.

The implementation is done without any extra dependencies to make it as generic as possible and without having to deal with whether or not I want to publish a docker image without and with the optional dependencies.

It's important that you FIRST run the migrations added in #1801 before moving to a remote solution if you intend to move your existing asset folder over.

It should be noted that I've only tested this with cloudflare R2 as that had an easy to setup free tier. I cannot vouch for any other service and I have only verified some basic operations. You'll have to look up for each service which values for region/bucket/url etc you need to provide. These can be added to the docs once known for a particular service.

**!! It should also be noted that by using a remote solution you accept the risk of data costs that could occur if something in PA behaves faulty and ends up requesting from the remote too much. !!**

This closes #1725 and #500

[^1]: The existing `assets.directory` config key is moved, it however didn't properly work as far as I'm aware so I don't think this should impact anybody. It now works properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,17 @@ tech changes will usually be stripped from release notes for the public
 
 ## Unreleased
 
+### Added
+
+-   [server] Option to store user uploaded assets in remote storage solutions (e.g. S3)
+
 ### Changed
 
 -   Variant shapes are completely reworked
     -   UI should be more intuitive
     -   No longer completely separate shapes with their own state
--   Asset size is now also stored in DB for easier user total size calculation
 -   [tech] DB storage of asset data is reworked
+    -   Asset size is now also stored in DB for easier user total size calculation
 
 ### Fixed
 
--- a/client/src/assets/utils.ts
+++ b/client/src/assets/utils.ts
@@ -1,23 +1,25 @@
 import { baseAdjust } from "../core/http";
+import { coreStore } from "../store/core";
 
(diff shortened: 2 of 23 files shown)


Kruptein:
This has now been updated in the docs for the next release.

See [here](https://deploy-preview-235--planarally.netlify.app/server/management/configuration/#subsection-storage). If the link no longer works, check the latest deploy of the [next release PR](https://github.com/Kruptein/planarally-docs/pull/235)

PR #1829, opened 2026-05-04
Global EventBus and Hook System

SuikaXhq:
These changes were implemented during the development of my mod. Given that this involves system refactoring, a thorough review would be greatly appreciated. Feedback and discussions are very welcome.

## Summary
This Pull Request refactors the PlanarAlly mod event system, moving away from a hardcoded, tightly-coupled pattern to a flexible, decoupled architecture using a global **EventBus** and **HookSystem**.

## Why the changes?
The legacy system required core systems (like `trackers` and `customData`) to manually iterate through all loaded mods to trigger events. This created several issues:
- **Tight Coupling**: Core systems had to know about the mod loading logic.
- **Rigid Interface**: Mods were limited to a fixed set of methods defined in the `Mod` object.
- **Maintenance Overhead**: Adding a new event required changes in multiple places (mod models, system index, and the mod-traversal logic).
- **No Data Interception**: There was no unified way for mods to synchronously intercept and modify data before it was processed.

### 1. EventBus (`core/eventBus.ts`)
A global singleton for asynchronous, notification-style events.
- **Usage**: Systems emit events (e.g., `eventBus.emit("tracker:added", data)`), and any interested party (mods or other systems) can subscribe.
- **Benefit**: Fire-and-forget logic that doesn't block the main execution flow.

### 2. HookSystem (`core/hooks.ts`)
A global singleton for synchronous data interception.
- **Usage**: Systems "pipe" data through hooks (e.g., `data = hooks.pipe("pre:tracker:update", data, context)`).
- **Benefit**: Allows mods to modify data (like HP changes or custom data fields) in real-time before it reaches the server or UI.

## Benefits for Internal Systems
Beyond decoupling mods, this architecture provides significant benefits for internal PlanarAlly development:
- **Clean Core Logic**: Core systems (e.g., `TrackerSystem`) no longer need to maintain complex lists of observers or manual callbacks. They simply emit events and continue their execution.
- **Cross-System Communication**: Systems like `AuraSystem` or `VisionSystem` can now react to tracker changes (e.g., a "Vision Range" tracker update) without the `TrackerSystem` needing to know about them.
- **Synchronous Data Pipelines**: The `HookSystem` allows for easy implementation of global rules or validation logic that intercepts data changes across all systems in a consistent way.

## Usage in Mods
Instead of implementing methods on the `Mod` class, mods now receive the `eventBus` and `hooks` instances during the `initGame` lifecycle.

### Example: Tracker & CustomData Synchronization
Mods can listen to events to keep their internal state in sync with PlanarAlly data:

```typescript
// 1. Reacting to a tracker change
api.eventBus.on("tracker:updated", ({ id, trackerId, delta }) => {
    console.log(`Shape ${id} tracker ${trackerId} changed:`, delta);
    // Update local mod state accordingly
});

// 2. Intercepting a customData update before it's processed
api.hooks.tap("pre:customData:update", (data, { id }) => {
    // Modify the data or add validation
    if (data.name === "secret_stat") {
        data.value = sanitize(data.value);
    }
    return data;
});
```

### My Example Mod: [Advanced Character Sheet](https://github.com/SuikaXhq/planarally-utility-mods/tree/main/packages/advanced-char-sheet)
The `advanced-char-sheet` mod uses both systems to maintain a two-way synchronization between PlanarAlly and its internal character data.

1. Using HookSystem for Interception: It hooks into tracker updates to sync HP/AC changes into its internal DataBlock:

```typescript
api.hooks.tap("pre:tracker:update", (delta, { id, tracker }) => {
    return SyncManager.handlePreTrackerUpdate(id, tracker, delta);
});
```

2. Using EventBus for Notifications: It listens to deletions and external updates to keep its mappings clean:

```typescript
// Reacting to a tracker being deleted externally
api.eventBus.on("tracker:removed", (payload) => {
    const globalId = api.getGlobalId(payload.id);
    const db = api.getDataBlock<any, CharSheetData>(...);
    if (db && db.data.trackerMappings.hp === payload.trackerId) {
        db.data.trackerMappings.hp = null;
        db.sync();
    }
});

// Reacting to customData updates from other sources
api.eventBus.on("customData:updated", (payload) => {
    const globalId = api.getGlobalId(payload.id);
    // Refresh character sheet UI if relevant data changed
});
```

## Key Changes
- Created `client/src/core/eventBus.ts` and `client/src/core/hooks.ts`.
- Refactored `CustomDataSystem` and `TrackerSystem` to use the new API.
- Removed legacy `mods.ts` files from `customData` and `trackers` directories.
- Updated `ModLoad` and `GameApi` types to expose the new systems to mods.
- Cleaned up the `initGame` injection logic in `client/src/mods/events.ts`.


client/src/core/eventBus.ts
```diff
@@ -0,0 +1,90 @@
...
+    // CustomData events
+    "customData:added": { id: LocalId; element: UiShapeCustomData; syncTo: Sync };
+    "customData:updated": { id: LocalId; elementId: ElementId; delta: Partial<ApiShapeCustomData>; syncTo: Sync };
+    "customData:removed": { id: LocalId; elementId: ElementId; syncTo: Sync };
+
+    // Tracker events
+    "tracker:added": { id: LocalId; tracker: Tracker; syncTo: Sync };
+    "tracker:updated": { id: LocalId; trackerId: TrackerId; delta: Partial<Tracker>; syncTo: Sync };
+    "tracker:removed": { id: LocalId; trackerId: TrackerId; syncTo: Sync };
+}
```
Kruptein: I would try to keep system specific events outside of this file.

We can make use of TS declaration merging and have this definition be empty

```ts
export interface EventMap {}
```

and add a new file in each relevant system (e.g. `mods.ts` or something)
```ts
import type { LocalId } from "../../../core/id";
import type { Sync } from "../../../core/models/types";

import type { Tracker, TrackerId } from "./models";

declare module "../../../core/eventBus" {
    interface EventMap {
        "tracker:added": { id: LocalId; tracker: Tracker; syncTo: Sync };
        "tracker:updated": { id: LocalId; trackerId: TrackerId; delta: Partial<Tracker>; syncTo: Sync };
        "tracker:removed": { id: LocalId; trackerId: TrackerId; syncTo: Sync };
    }
}
```

This removes the need of adding system specific concepts into the core event bus logic.

(The same applies to the hooks)

client/src/core/eventBus.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    private off<K extends keyof EventMap>(event: K, handler: EventHandler<K>): void {
+        const handlers = this.listeners[event];
+        if (!handlers) return;
+        const idx = handlers.indexOf(handler);
+        if (idx !== -1) handlers.splice(idx, 1);
+    }
+}
+
+export const eventBus = new EventBus();
```
Kruptein: _I'm aware that the comment I'm about to make is very hypocritical as the existing codebase uses this pattern a lot. It's something I'm not happy with though, hence the remark._

Unless you actually envision a use for multiple EventBusses co-existing, it might be cleaner to skip the class all together and instead do something like

```ts
export interface EventMap {}

type EventHandler<K extends keyof EventMap> = (payload: EventMap[K]) => void | Promise<void>;

const listeners: { [K in keyof EventMap]?: EventHandler<K>[] } = {};

// ...

function once<K extends keyof EventMap>(event: K, handler: EventHandler<K>): () => void {
    // ...
}

function emit<K extends keyof EventMap>(event: K, payload: EventMap[K]): void {
    // ...
}

export const eventBus = { on, once, emit };
```

client/src/game/systems/customData/index.ts
```diff
@@ -125,7 +127,12 @@ class CustomDataSystem implements ShapeSystem<UiShapeCustomData[]> {
         const target = ($.leases.has(id) ? $.data.get(id) : mutable.data.get(id)) ?? [];
         const element = target.find((element) => element.id === elementId);
         if (element === undefined) return;
+
+        const syncTo = { server: false, ui: true };
+        data = hooks.pipe("pre:customData:update", data, { id, element, syncTo }) as ApiShapeCustomData;
```
Kruptein: the cast here feels dangerous, a faulty mod implementation could easily return a Partial version of the type. [^1]

Instead I would change the hook definition to be

```ts
pipe<K extends keyof HookMap, V extends HookMap[K]["value"]>(
        hook: K,
        initialValue: V,
        context: HookMap[K]["args"],
    ): V

// combined with
    "pre:customData:update": {
        args: { id: LocalId; element: UiShapeCustomData; syncTo: Sync };
        value: Partial<ApiShapeCustomData> | ApiShapeCustomData;
    };
```

(You'll have to update some other things as well)

I haven't tested the (mod) implementation side of this, so if this is unwieldy, I would maybe just offer 2 different hooks here.

[^1]: granted, a faulty mod that is pure JS can always return garbage data, but the point is more about the TS side appearing OK even though it could potentially be wrong.

Kruptein (changes requested):
Thanks for tackling this!

While you certainly have a lot of valid points I do feel that some just moved location. e.g. "Mods were limited to a fixed set of methods defined in the Mod object." Instead of being limited to what is defined on the Mod object, you're now limited to what's defined on the event-bus/hook-system.

Similarly by going through the event system, you're not removing coupling between systems in my opinion but just hiding it.

That said I don't dislike the proposed changes and think they do improve the API.

The hook system is a clear improvement to ensure that these style of calls are implemented in a consistent manner.

For the events implemented in the event-bus I was initially a bit conflicted as that data was already available using vue watchers on the system state. Thinking on it further however I do realise that this requires knowledge on how vue reactivity works and for some mods that might be overkill. So I think I'm ok with the events added and power-users can still use watchers where they see fit.

So overall great job :)

client/src/game/systems/customData/types.ts
```diff
@@ -69,3 +70,20 @@ export const customDataKindMap: CustomDataKindMap = {
...
+}
+
+declare module "../../../core/hooks" {
+    interface HookMap {
+        "pre:customData:update": {
+            args: { id: LocalId; element: UiShapeCustomData; syncTo: Sync };
+            value: Partial<ApiShapeCustomData> | ApiShapeCustomData;
+        };
+    }
+}
```
SuikaXhq, line 89: I think these declarations should be in a `models.ts` similar to what have been done in Tracker system. But considering that refactoring this file to 2 files (`types` and `models`) would lead to more importing changes in other files, I decided to leave them here for now.

client/src/core/eventBus.ts
SuikaXhq: agree with you.

client/src/core/eventBus.ts
SuikaXhq: I actually referred to the other systems and found they are singletons, so I implemented it also with singleton XD. Anyway I refactored it.

client/src/game/systems/customData/index.ts
SuikaXhq: You're correct.

Offering 2 different hooks may bring in more complexity here so I think one is fine.

client/src/core/eventBus.ts
Kruptein: yeah it's why I prefaced it with the "hypocritical" comment as I know it's prevalent :p

client/src/game/systems/customData/types.ts
Kruptein: I agree with this, having both types and models feels a bit strange as they convey the same meaning. I'm fine with doing it in this PR or in separate PR.

Kruptein (approved):

PR #1830, opened 2026-05-04
feat: add modals and activateTool API to mod-api

SuikaXhq:
Exposed `activateTool` to the mod API, enabling interaction with tools.

Added modals API for mods to trigger standard dialogs, maintaining UI consistency.

### DIFF
--- a/client/src/core/plugins/modals/plugin.ts
+++ b/client/src/core/plugins/modals/plugin.ts
@@ -13,7 +13,7 @@ import type { SelectionBoxFunction } from "./selectionBox";
 
 const modalSymbol: InjectionKey<Modals> = Symbol("PlanarAllyModals");
 
-interface Modals {
+export interface Modals {
     confirm: ConfirmFunction;
     prompt: PromptFunction;
     selectionBox: SelectionBoxFunction;
@@ -42,17 +42,23 @@ async function createModals(): Promise<Modals> {
     };
 }
 
-export const modals = shallowReactive<Partial<Modals>>({});
+const _modals = shallowReactive<Partial<Modals>>({});
+
+export const modals: Modals = {
+    confirm: (...args) => _modals.confirm!(...args),
+    prompt: (...args) => _modals.prompt!(...args),
+    selectionBox: (...args) => _modals.selectionBox!(...args),
+};
 
 export const PlanarAllyModalsPlugin: Plugin = async (App) => {
-    Object.assign(modals, await createModals());
-    App.provide(modalSymbol, modals as Modals);
+    Object.assign(_modals, await createModals());
+    App.provide(modalSymbol, modals);
 };
 
 export function useModal(): Modals {
-    const _modals = inject(modalSymbol);
-    if (_modals === undefined) {
+    const modalsInstance = inject(modalSymbol);
+    if (modalsInstance === undefined) {
         throw new Error("Could not inject modals");
     }
-    return _modals;
+    return modalsInstance;
 }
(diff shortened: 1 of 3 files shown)


client/src/core/plugins/modals/plugin.ts
```diff
@@ -42,15 +42,23 @@ async function createModals(): Promise<Modals> {
...
+    confirm: (...args) => _modals.confirm!(...args),
+    prompt: (...args) => _modals.prompt!(...args),
+    selectionBox: (...args) => _modals.selectionBox!(...args),
+};
+
 export const PlanarAllyModalsPlugin: Plugin = async (App) => {
-    const modals = await createModals();
+    Object.assign(_modals, await createModals());
     App.provide(modalSymbol, modals);
 };
```
Kruptein: Is there a reason why a shallow reactive is used?

Couldn't this be simplified to the following:

```suggestion
export let modals: Modals;

export const PlanarAllyModalsPlugin: Plugin = async (App) => {
    modals = await createModals();
    App.provide(modalSymbol, modals);
};
```

client/src/mods/events.ts
```diff
@@ -6,25 +6,29 @@ import { getDataBlockFunctions } from "./db";
...
                 mod.events?.initGame?.({
                     systems: SYSTEMS,
                     systemsState: SYSTEMS_STATE,
-                    ui,
+                    ui: {
+                        shape: {
+                            registerContextMenuEntry,
+                            registerTab,
+                        },
+                        activateTool,
```
Kruptein: I can see the argument of adding this to the ui section, but I feel like conceptually it might fit better outside of it.

You're not activating a tool with the intent of interacting with some UI element, but rather to enable a particular functionality, if that makes sense :p

Kruptein (changes requested):

client/src/mods/events.ts
SuikaXhq: I've thought `activateTool` is a "ui" function as it only changes the current active tool. Maybe it should be placed in a new section named like `gamePlay` (indicates interacting with in-game functionalities)?

client/src/core/plugins/modals/plugin.ts
SuikaXhq: The [check](https://github.com/Kruptein/PlanarAlly/pull/1830/changes/6ea32247004095e17fa42dfc2ddd4be7ec3bf794) asked not to export let variables, but it seems turned out to be an ad of SonarCloud?

Kruptein: The behaviour of that button is odd, it's not an ad. The URL mentioned in the comment actually leads to a real report https://sonarcloud.io/project/issues?id=Kruptein_PlanarAlly&issues=AZ4Su2ykHiNwq88h1lUS&open=AZ4Su2ykHiNwq88h1lUS&pullRequest=1830 It just so happens that the button expects the user to be logged in, which is not going to be the case for contributors.

I don't really agree with sonar in this instance. Or rather the alternatives are too cumbersome (e.g. you have to repeat a bunch in your changed approach) and realistically in es6 you can't re-bind an exported let variable. So I'll just suppress the error in sonar if it triggers again.

client/src/mods/events.ts
Kruptein: Sure gameplay is fine for now. This is an area where some breaking changes are likely in the future as we discover which parts of PA people want to interact with.

Kruptein:
there seems to be a merge conflict

SuikaXhq:
conflict was resolved. Ready for merge now

Issue #984, opened 2022-06-23
[FEAT] The trackers can be be adjusted up or down


TOUHOUCON:
After player level up, it is messy to add resources on the tracker.
![2022-06-23_124321](https://user-images.githubusercontent.com/49478304/175212371-7b7edd63-ec2d-46af-a3a4-e729417b0735.jpg)
I hope to adjust the existing resources up and down

10leej:
You want the ability to reorder the trackers right?

TOUHOUCON:
> 您想要重新排序跟踪器的能力吗？

YES！

PR #1866, opened 2026-09-17
[FEAT] Move order of trackers on a given shape

AlkalineSalts:
This PR adds the ability to move the trackers on a particular shape upwards or downwards in the list of that shape's trackers. It does this by essentially swapping what values belong to which UUID (x,y -> y,x). This is intended to add feature request #984. 

The changes as presented have do not account for either race conditions or partial updates. 

Race conditions:

If a client moves a tracker for a shape up or down and a second client receives one Shape.Options.Tracker.Update from the server but doesn't receive the second before trying to move the shape up or down, one of the trackers will have their values overwritten (by values I mean all fields of the tracker class besides the UUID) (x,y -> x,x). 

This could potentially be fixed client side, by checking if adjacent trackers have the same values and preventing any movement of trackers if that is the case. 

Partial updates:

Since the client must send two separate Shape.Options.Tracker.Update messages in order to properly complete the operation, if transmission gets interrupted after the first but before the second has completed, than the same as in the previous issue will occur and one of the tracker's will have their values overwritten (x,y -> x,x).

This can't be resolved client side, would require changes to server side code to address.


Kruptein:
Thanks for contributing!

Swapping uuids gets the job done, but has multiple downsides that end up making the solution not desirable.

There are the race conditions and partial updates you've mentioned yourself, but there are two other things that I can think off the top of my head:

- uuids are to a certain degree a contract that the ID can be used to talk about 1 specific thing, this means anything that stores a tracker's uuid, should be able to assume that this is a stable reference. While PA itself currently does not have things that actively refer to trackers, a future update or a mod could easily retain a uuid for a particular purpose. Re-ordering trackers by swapping uuids, will have unpredictable consequences for these scenarios.
- if at a later point this UI were to change and allow for drag and drop behaviour, where a tracker can be moved multiple positions down at once, we have to do a shuffle of multiple uuids, which increases the issues you mentioned yourself.

I'm not super worried about race conditions, as ultimately tracker updates are relatively infrequent, but I do value the uuid stability highly. So instead I would suggest to add an explicit index column to the tracker database model and use an explicit new socket event to reorder trackers on a shape. This extra index is cheap and does the work without any race conditions or partial update issues and keeps the uuid reference stable.

AlkalineSalts:
Alright, I'll add in a new unique int column, maybe "positioning" or "ordering" or something like that and return trackers in an ordering based on the column. Swapping will be done by swapping this column's value in the DB and by swapping trackers positions in their respective arrays on the client. Only thing I'm not sure I'll figure out is how to migrate correct values into existing sqlite databases, but I'll ask if I can't figure it out.

AlkalineSalts:
Should a new ordering column for the tracker table in sqlite be transmitted to clients? I'm in favor of that information only being on the server, but I could see either way being viable approaches.

